PR #54, opened 2021-07-20
CPP-727 add eslint import plugin recommended config

apaleslimghost:
we've got the plugin installed but aren't actually using any of its rules! this would have caught #53. this has a fair chance of causing some linting errors in CI which i'll fix in this branch if it does (writing this commit on github bc easier)


ivomurrell (approved):
No recollection as to why I didn't approve this the first time round 😅

PR #131, opened 2021-10-12
CPP-620 add basic node plugin for running node things

apaleslimghost:
adds a plugin to run a node app directly on the `run:local` hook. depends on #130. plugin is deliberately unconfigurable (apart from server entry point) for now. it's easy to add options but impossible to delete them.

### DIFF
--- a/packages/sandbox/package.json
+++ b/packages/sandbox/package.json
@@ -5,7 +5,8 @@
   "scripts": {
     "test": "dotcom-tool-kit test:local",
     "build": "dotcom-tool-kit build:local",
-    "heroku-postbuild": "dotcom-tool-kit release:remote build:remote cleanup:remote"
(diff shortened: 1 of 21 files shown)


packages/sandbox/package.json
```diff
@@ -24,5 +25,8 @@
   },
   "volta": {
     "extends": "../../package.json"
+  },
+  "dependencies": {
+    "@dotcom-tool-kit/node": "file:../node"
```
ivomurrell, line 30: **issue:** I think we need `dotcom-tool-kit` installed as a dependency now that we're using the `run:local` hook at runtime

ivomurrell: why isn't the `node` package a dev dependency then? 😛

apaleslimghost: be...cause... i forgot to add `--save-dev` when i did this

PR #142, opened 2021-10-26
Integrate change API orb into tool kit orb

ivomurrell:
This will post any heroku deployments in CircleCI to the [Change API](https://github.com/Financial-Times/change-api) so changes in services can be tracked.

### DIFF
--- a/orb/src/jobs/heroku-promote.yml
+++ b/orb/src/jobs/heroku-promote.yml
@@ -2,6 +2,16 @@ parameters:
   node-version:
     default: "12.22"
     type: string
+  systemCode:
+    default: $CIRCLE_PROJECT_REPONAME
+    description: >-
+      The systemcode of the system being changed. Defaults to the repository
+      name.
+    type: string
+  environment:
+    default: unknown
+    description: The environment in which the system is being changed.
+    type: string
 
 executor:
   name: node/default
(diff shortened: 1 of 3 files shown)


orb/src/jobs/heroku-promote.yml
```diff
@@ -12,3 +22,5 @@ steps:
   - run:
       name: Deploy to production
       command: npx dotcom-tool-kit deploy:production
+  - change-api/generate:
+      systemCode: <<parameters.node-version>>
```
leannecornish-ft, line 26: Suggestion: I think this should be systemCode?
 ```suggestion
       systemCode: <<parameters.systemCode>>

orb/src/jobs/heroku-promote.yml
```diff
@@ -12,3 +22,5 @@ steps:
   - run:
       name: Deploy to production
       command: npx dotcom-tool-kit deploy:production
+  - change-api/generate:
```
leannecornish-ft, line 25: Question: it looks like the `change-api/generate` also uses `changeApiKey`, `user`, `environment`, `slackChannels` and `extraProperties` - is there a reason why they're not included here?

leannecornish-ft (commented):
Looks good! Just one question and possibly a typo...

orb/src/jobs/heroku-promote.yml
ivomurrell: The defaults seem fine to me so didn't think we needed to surface them as options – we can add them later if people need them

leannecornish-ft (approved):

leannecornish-ft (approved):

PR #144, opened 2021-11-01
CPP-637 Add guide to migrate away from n-gage

ivomurrell:


docs/migrating-to-tool-kit.md
```diff
@@ -0,0 +1,181 @@
...
+suggestions and the migration tool doesn't perform any actions so you'll need to
+delete the Makefile yourself if you don't think it's needed anymore. Remember
+that – assuming you didn't delete it in the earlier step – some of the `make`
+commands come from `n-gage` itself and won't necessarily be in the Makefile,
+such as `make install`.
+
+## Migrating Your Heroku Pipeline
+
+One thing that the migration tool cannot automate is reconfiguring your Heroku
+pipeline to integrate properly with Tool Kit. You will need to ensure that your
```
serena97, line 175: **suggestion:** perhaps we can add here, that

On top of installing the plugin `circleci-heroku`, which provides hooks that circleci runs such as 'deploy:review' that creates a review app,  you'll have to ensure that your
pipeline is connected to the project's ...

just to be clear that you'll still need the heroku plugin as well as following the step by step guide

serena97 (approved):
I've left a few suggestions below but it looks good to me! I really like how you structured it step by step explaining each question from the create script

docs/migrating-to-tool-kit.md
ivomurrell: The `front-end` plugin will install `circleci-heroku` automatically so I'm not sure if we need to be explicit about it?

serena97: ah right of course, I agree - we don't have to be explicit about it then

PR #145, opened 2021-11-03
CPP-638 add circleci nightly workflow to task

emortong:
[Ticket CPP-638](https://financialtimes.atlassian.net/browse/CPP-638)

This PR modifies the circleci plugin so that it:

- errors and asks user to recreate if `nightly` workflow is not present
- adds optional `addToNightly` property to the `CircleConfig` interface
- if `addToNightly` is set to **true** it will clone the `tool-kit` job, remove the `waiting-for-approval` requirement and add it to the `nightly` workflow
- sets the `addToNightly` property to **true** for `BuildCI` and `TestCI`
- fixes tests

### DIFF
--- a/core/cli/test/__snapshots__/index.test.ts.snap
+++ b/core/cli/test/__snapshots__/index.test.ts.snap
@@ -630,6 +630,7 @@ Object {
   },
   "hooks": Object {
     "build:ci": BuildCI {
+      "addToNightly": true,
       "circleConfigPath": ".circleci/config.yml",
       "id": "build:ci",
       "job": "tool-kit/build",
@@ -645,6 +646,7 @@ Object {
         "hooks": Object {
           "build:ci": [Circular],
           "test:ci": TestCI {
+            "addToNightly": true,
             "circleConfigPath": ".circleci/config.yml",
             "id": "test:ci",
             "job": "tool-kit/test",
@@ -1858,6 +1860,7 @@ Object {
       },
     },
     "test:ci": TestCI {
+      "addToNightly": true,
       "circleConfigPath": ".circleci/config.yml",
       "id": "test:ci",
       "job": "tool-kit/test",
@@ -1871,6 +1874,7 @@ Object {
         "TestCI": [Function],
         "hooks": Object {
           "build:ci": BuildCI {
+            "addToNightly": true,
             "circleConfigPath": ".circleci/config.yml",
             "id": "build:ci",
             "job": "tool-kit/build",
@@ -2583,6 +2587,7 @@ Object {
       "TestCI": [Function],
       "hooks": Object {
         "build:ci": BuildCI {
+          "addToNightly": true,
           "circleConfigPath": ".circleci/config.yml",
           "id": "build:ci",
           "job": "tool-kit/build",
@@ -2595,6 +2600,7 @@ Object {
           "plugin": [Circular],
         },
         "test:ci": TestCI {
+          "addToNightly": true,
           "circleConfigPath": ".circleci/config.yml",
           "id": "test:ci",
           "job": "tool-kit/test",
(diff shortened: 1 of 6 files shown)


packages/circleci/src/circleci-config.ts
```diff
@@ -137,7 +156,7 @@ export default abstract class CircleCiConfigHook {
       config.orbs['tool-kit'] = `financial-times/dotcom-tool-kit@${currentVersion}`
     }
 
-    if (!(config.workflows?.['tool-kit'] as Workflow).jobs) {
+    if (!(config.workflows?.['tool-kit'] as Workflow).jobs || !(config.workflows?.['nightly'] as Workflow).jobs) {
```
ivomurrell, line 159: **issue:** shouldn't this be checking that both of the workflows are valid, not either?
```suggestion
    if (!(config.workflows?.['tool-kit'] as Workflow).jobs && !(config.workflows?.['nightly'] as Workflow).jobs) {
```

packages/circleci/src/circleci-config.ts
```diff
@@ -146,10 +165,18 @@ export default abstract class CircleCiConfigHook {
...
     const jobs = (config.workflows!['tool-kit'] as Workflow).jobs!
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const nightlyJobs = (config.workflows!['nightly'] as Workflow).jobs!
     const job = this.jobOptions ? { [this.job]: this.jobOptions } : this.job
     // Avoid duplicating jobs (this can happen when check() fails when the version is wrong)
     if (!jobs.some((candidateJob) => isEqual(candidateJob, job))) {
       jobs.push(job)
+      if (this.addToNightly) {
+        // clone job and remove waiting for approvals
+        const clonedJob = JSON.parse(JSON.stringify(job))
```
ivomurrell, line 176: **nitpick:** this encoding/decoding can be lossy, so I usually prefer using [`lodash.clonedeep`](https://lodash.com/docs/4.17.15#cloneDeep). not a big deal in this case though

ivomurrell (changes requested):
**issue:** we are currently not fixing a config if it has the `tool-kit` workflow but not the `nightly` workflow, as the `check` method only checks for the former, so gives a false positive and the `install` method skipped. would we be able to check that the nightly workflow exists in the `check` method so that the correct config will be generated if you moved to this version?

a couple of other small issues but I think this is a good solution to the issue! (sorry for having to deal with my hacky code... 😰)

packages/circleci/src/circleci-config.ts
next-team: nice catch!! thank you!

emortong: nice catch!! thank you!

leannecornish-ft:
@ivomurrell  looks like this has some failing tests? 
also praise: really helpful extended commit messages, thanks!

packages/circleci/src/circleci-config.ts
```diff
@@ -137,20 +150,36 @@ export default abstract class CircleCiConfigHook {
...
     // Avoid duplicating jobs (this can happen when check() fails when the version is wrong)
     if (!jobs.some((candidateJob) => isEqual(candidateJob, job))) {
       jobs.push(job)
     }
+    if (this.addToNightly && !nightlyJobs.some((candidateJob) => isEqual(candidateJob, job))) {
+      const nightlyJob = this.jobOptions
+        ? {
+            [this.job]: {
+              ...this.jobOptions,
+              requires: this.jobOptions.requires?.filter((x: string) => x !== 'waiting-for-approval')
```
serena97, line 177: **praise**: great edge case to account for

serena97 (approved):
looks good to me!

ivomurrell (approved):
I tidied up the hacky code this PR is based on so should now be ready to merge and will hopefully be easier to refactor in the future.

emortong:
@ivomurrell thank you for helping with this Ivo!!

PR #146, opened 2021-11-03
CPP-608 Validate that plugins are valid when loading them

ivomurrell:
We do this by defining an abstract `Hook` class that all hooks must be derived from. This means we can now check all the types on packages we load have the appropriate `instanceof`, enforcing a structure for plugins that should avoid us running into hard-to-debug issues down the road.

Any ideas for better names for the 'hook' package are welcome: it defines both the `Hook` and `Plugin` types as they are mutually dependent!


serena97 (approved):
It looks good to us! One thing we can do is move the code in hook/src/types.ts into the types/src/index.ts as the Plugin type belongs to the types package more.

serena97 (approved):
Looks good to @leannecornish-ft and I! Makes sense that the tasks package is now removed and the things inside it are moved to types package to avoid circular dependencies. 

Also it's great that you have included detailed commit messages, it was helpful to go through the PR via commits 👍

PR #147, opened 2021-11-05
Local vault fix CPP-651

leannecornish-ft:
When `dotcom-tool-kit run:local` runs it uses the `vault` plugin to get the required environment variables for `next-router` and the app being developed. It needs an auth token to gain access, and first checks for an existing one in `~/.vault-token`. However, these tokens expire after 8 hours, and it's not checking whether this is a current token so fails with an authentication failure if it's using an expired token. 

This PR adds a check for an expired token, and requests a new one if found. It also writes any new tokens to file.

### DIFF
--- a/packages/vault/src/index.ts
+++ b/packages/vault/src/index.ts
@@ -2,6 +2,7 @@ import fetch from '@financial-times/n-fetch'
 import { promises as fs } from 'fs'
 import path from 'path'
 import os from 'os'
+import moment from 'moment'
 import { ToolKitError } from '@dotcom-tool-kit/error'
 import { getOptions } from '@dotcom-tool-kit/options'
 import { VaultOptions } from '@dotcom-tool-kit/types/lib/schema/vault'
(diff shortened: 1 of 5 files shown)


packages/vault/src/index.ts
```diff
@@ -81,29 +82,45 @@ export class VaultEnvVars {
...
+      const vaultTokenFile = path.join(os.homedir(), '.vault-token')
       try {
-        const vaultTokenFile = await fs.readFile(path.join(os.homedir(), '.vault-token'), {
-          encoding: 'utf8'
-        })
-        if (vaultTokenFile) {
-          return vaultTokenFile
-        } else if (VAULT_AUTH_GITHUB_TOKEN) {
+        const stats = await fs.stat(vaultTokenFile)
+        const fileExpired = moment().diff(stats.birthtime, 'hours') > 8
```
serena97, line 88: **nitpick**: while the logic makes sense, it might make sense to pull the magic number 8 from a config file or api if we can just in case the token expiry time changes

packages/vault/test/index.test.ts
```diff
@@ -0,0 +1,112 @@
+import { describe, it, beforeAll, beforeEach, afterAll, jest, expect } from '@jest/globals'
+import { VaultEnvVars } from '../src/index'
+import fetch from '@financial-times/n-fetch'
+import { mocked } from 'ts-jest/utils'
+import fs from 'fs'
+process.env.CIRCLECI = 'true'
```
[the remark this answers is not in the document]
leannecornish-ft: No, this was left over from me testing locally and should be removed, good spot!

packages/vault/src/index.ts
```diff
@@ -58,14 +58,62 @@ export class VaultEnvVars {
...
+      const vaultToken = await fs.readFile(this.vaultTokenFile, {
+        encoding: 'utf8'
+      })
+      console.log('testing current token')
+      const validToken = await this.fetchTest(vaultToken)
+      if (validToken) {
+        console.log('success!')
+        return vaultToken
+      } else {
+        throw new ToolKitError('current token invalid, requesting new one...')
```
serena97, line 92: **suggestion:** if line 92 or line 95 executes, I'm not sure if their error would be output as it might be swallowed by the catch block in line 135. For instance, if line 92 executes and therefore `new ToolKitError('current token invalid, requesting new one...')` is thrown, and then in line 136 `VAULT_AUTH_GITHUB_TOKEN` is undefined, we will end up with the error 'VAULT_AUTH_GITHUB_TOKEN variable is not set' with the 'current token invalid, requesting new one...'. being swallowed. To guarantee the user can see the error from line 92 / 95, I think we could log out the error here as well

```suggestion
        const message = 'current token invalid, requesting new one...'
        console.error(message)
        throw new ToolKitError(message)
```

packages/vault/src/index.ts
```diff
@@ -82,28 +130,21 @@ export class VaultEnvVars {
...
-            method: 'POST',
-            headers: { 'Content-type': 'application/json' },
-            body: JSON.stringify({ token: VAULT_AUTH_GITHUB_TOKEN })
-          })
-          return token.auth.client_token
+        const token = await this.handleLocalToken()
+        return token
+      } catch {
+        if (VAULT_AUTH_GITHUB_TOKEN) {
+          try {
```
serena97, line 137: **suggestion:** I don't think we need to wrap line 138 to 139 in a try catch as the catch block is throwing the same error that line 138 would throw

serena97 (changes requested):
It is looking better now with this mechanism, where the code attempts to fetch the token to see if it's still valid, and if not it gets the token from vault. I'm happy to approve it once the minor suggestions to do with error handling are addressed

serena97 (approved):

PR #151, opened 2021-11-24
Vault options fix

leannecornish-ft:
A bug that was introduced in when we moved the vault options into the [vault plugin](https://github.com/Financial-Times/dotcom-tool-kit/blob/51fb03f3035413d91556b08b9af1218ab4facd5c/packages/vault/src/index.ts#L42). Unfortunately, this is a non-standard plugin that doesn't contain a task to get its options from installation, when [`getOptions`](https://github.com/Financial-Times/dotcom-tool-kit/blob/51fb03f3035413d91556b08b9af1218ab4facd5c/packages/options/src/index.ts#L5) is called. Instead, it calls `getOptions` when the `getEnvVars` class is instantiated by another plugin, and at this point it returns an empty object, presumably because `options` is no longer in global memory.

This PR adds a function in `vault` that reads the .toolkitrc.yml of the consuming app, and returns the relevant options for `vault`. This is the simplest way I can think, but very happy to hear different approaches 🙂

### DIFF
--- a/packages/vault/src/getVaultOptions.ts
+++ b/packages/vault/src/getVaultOptions.ts
@@ -1,0 +1,17 @@
+import { VaultOptions } from '@dotcom-tool-kit/types/lib/schema/vault'
+import { ToolKitError } from '@dotcom-tool-kit/error'
+import { cosmiconfig } from 'cosmiconfig'
+
+const explorer = cosmiconfig('toolkit', { ignoreEmptySearchPlaces: false })
+
+export async function getVaultOptions(): Promise<VaultOptions> {
+  try {
+    const result = await explorer.search(process.cwd())
+    const { team, app } = result?.config.options['@dotcom-tool-kit/vault']
+    return { team, app }
+  } catch {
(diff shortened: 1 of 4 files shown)


packages/vault/src/getVaultOptions.ts
```diff
@@ -0,0 +1,17 @@
+import { VaultOptions } from '@dotcom-tool-kit/types/lib/schema/vault'
+import { ToolKitError } from '@dotcom-tool-kit/error'
+import { cosmiconfig } from 'cosmiconfig'
```
apaleslimghost, line 3: **thought**: it doesn't feel right doing it this way. this doesn't follow the same behaviour as `cli` for loading and merging option defaults from every plugin, and having a separate implementation of this in a single plugin is almost certainly going to cause inconsistencies down the line. if the support for non-`Task` plugins for loading options via `cli` isn't working, we should fix that rather than implementing it again.

leannecornish-ft:
User error 😅 , the cli package hadn't been updated to 1.2 from beta.

PR #152, opened 2021-11-24
CPP-654 Organise packages into appropriate directories

ivomurrell:
- Now all packages which define Tool Kit plugins, i.e., packages which export hooks or tasks and are consumed by other plugins or apps in their `plugins` fields are located in the `plugins` directory.
- All libraries that other packages pull in but are not Tool Kit plugins are located in the `lib` directory.
- All other packages remain in the `packages` directory.

These changes will help us quickly differentiate between plugins and helper libraries, both as developers and to allow us to run integration tests on only the plugins.


tsconfig.json
```diff
@@ -2,94 +2,94 @@
...
     {
-      "path": "packages/heroku"
+      "path": "plugins/heroku"
     },
     {
-      "path": "packages/upload-assets-to-s3"
+      "path": "plugins/upload-assets-to-s3"
     },
     {
       "path": "packages/cli"
```
apaleslimghost, line 35: **suggestion**: i think `core` might be a clearer name than `packages`?

ivomurrell: would you say the sandbox is 'core'?

apaleslimghost: `sandbox` is its own category really. maybe it should actually become an integration test fixture that we can also use for manual testing?

ivomurrell: that sounds reasonable. I'll leave it in `core` for this PR as it feels like that requires a bit more thought if that's okay

leannecornish-ft (approved):
LGTM - the re-structuring of the packages makes sense, I like the name `core`, and it builds successfully so you have presumably found all of the places to copy/paste 😄 

Praise: great use of `git mv` (or whatever you did!) - it makes reviewing this 270-file change a lot easier!

serena97 (approved):
looks good to me too! one thing that we should also do is update the list of plugins in `lib/types/test/plugin.test.js`,  but it might make sense to do it as a new subtask and separate PR

ivomurrell:
> great use of `git mv` (or whatever you did!) - it makes reviewing this 270-file change a lot easier!

I did indeed make sure to use `git mv` (maybe it would have worked fine regardless...?) though it did get a bit confused about some of the empty `.toolkitrc.yml` files after I rebased onto `main` unfortunately!

> one thing that we should also do is update the list of plugins in `lib/types/test/plugin.test.js`

ah I'll do that before merging, I only didn't do it already because I originally wrote this before #146 was merged

lib/types/test/plugin.test.js
```diff
@@ -4,34 +4,19 @@
...
-])('%s integration test', (plugin) => {
-  const packagePath = path.join(__dirname, `../../${plugin}`)
+const pluginDir = path.join(__dirname, '../../../plugins')
+
+function getPlugins() {
+  const pluginDirContents = fs.readdirSync(pluginDir, { withFileTypes: true })
+  return pluginDirContents.filter((plugin) => plugin.isDirectory()).map((plugin) => plugin.name)
+}
+
+describe.each(getPlugins())('%s integration test', (plugin) => {
```
serena97, line 18: that's neat! 👍

serena97 (commented):
the refactoring looks good to me! I think we should also add this integration test to the pipeline  after the unit tests are run

**edit**: sorry, I was searching for the wrong keyword in the pipeline so missed it - you are right, `lib/types/test/plugin.test.js` is run in run package tests!

ivomurrell:
> I think we should also add this integration test to the pipeline after the unit tests are run

It already runs when `npm test` is called along with the unit tests

PR #153, opened 2021-12-01
Don't run setup orb job until workflow is approved

ivomurrell:
During a nori run I noticed that currently `tool-kit/setup` does not require the `waiting-for-approval` job that optionally runs for nori branches, which instead only blocks the `tool-kit/build` job and onwards. This doesn't make much sense seeing as the `setup` job runs one of the most expensive operations in the CircleCI workflow – the `npm install` step – so really shouldn't be running until approved in order to reduce the load on the CI when we push a series of nori commits.


apaleslimghost (approved):
👀 👀 👀 👀

apaleslimghost (approved):

PR #157, opened 2022-01-05
fix husky config format installed by hook

apaleslimghost:
also add a `peerDependency` on the correct version of husky to make setting it up easier

### DIFF
--- a/lib/husky-hook/src/index.ts
+++ b/lib/husky-hook/src/index.ts
@@ -1,7 +1,9 @@
 import { PackageJsonHelper } from '@dotcom-tool-kit/hook'
 
 type HuskyHooks = {
-  [hook: string]: string
+  hooks: {
+    [hook: string]: string
+  }
 }
 
 export abstract class HuskyHook extends PackageJsonHelper {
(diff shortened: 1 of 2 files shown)


lib/husky-hook/src/index.ts
```diff
@@ -10,11 +12,11 @@ export abstract class HuskyHook extends PackageJsonHelper {
   async install(): Promise<void> {
     let command = `dotcom-tool-kit ${this.hook}`
     const huskyHooks = this.packageJson.getField<HuskyHooks>(this.field) || {}
-    const existingCommand = huskyHooks[this.key]
+    const existingCommand = huskyHooks.hooks[this.key]
```
apaleslimghost, line 15: gah, this won't work if the field isn't there in the first place

plugins/husky-npm/package.json
```diff
@@ -12,6 +12,9 @@
   "dependencies": {
     "@dotcom-tool-kit/husky-hook": "file:../../lib/husky-hook"
   },
+  "peerDependencies": {
+    "husky": "^4.3.8"
```
ivomurrell, line 16: **issue**: We want `peerDependencies` to be as permissive as possible, so older minor/patch versions should be allowed too. In theory, we might also be able to support some other major versions if they read the config in a similar way but annoyingly the migration guides are overwritten for each new major version so the compatibility is not immediately obvious.

```suggestion
    "husky": "4.x"
```

ivomurrell (changes requested):
Thanks for spotting this! I have no idea how I missed it when testing?!

**issue**: the `check` method will probably also have to be overridden from `PackageJsonHelper` so that it correctly checks `husky.hook` rather than just `husky` for the tool kit hook.

ivomurrell (approved):
LGTM!

PR #158, opened 2022-01-06
Fix ignored lint-staged-npm glob options

ivomurrell:
`lint-staged-npm` allows you to set the globs used by lint-staged in your `.toolkitrc.yml` config. Previously, it would read the options when the class was constructed. However, now that we construct the classes when first loading the plugins (in order to verify they're instances of the `Hook` class) this code is now running before the options from the config file have been set. Instead, lazily read the options the first time the key property is accessed, which should be just after the place where they were previously being constructed.

This should fix all `lint-staged-npm` hooks being assigned to the default `**/*.js` glob regardless of their `formatGlob` and `testGlob` options.

### DIFF
--- a/plugins/lint-staged-npm/src/index.ts
+++ b/plugins/lint-staged-npm/src/index.ts
@@ -4,14 +4,20 @@ import { getOptions } from '@dotcom-tool-kit/options'
 class TestStaged extends LintStagedHook {
   static description = 'format prettier'
 
-  key = getOptions('@dotcom-tool-kit/lint-staged-npm')?.testGlob ?? '**/*.js'
+  _key?: string
+  get key(): string {
+    return (this._key ??= getOptions('@dotcom-tool-kit/lint-staged-npm')?.testGlob ?? '**/*.js')
+  }
   hook = 'test:staged'
 }
 
 class FormatStaged extends LintStagedHook {
   static description = 'format prettier'
 
-  key = getOptions('@dotcom-tool-kit/lint-staged-npm')?.formatGlob ?? '**/*.js'
+  _key?: string
+  get key(): string {
+    return (this._key ??= getOptions('@dotcom-tool-kit/lint-staged-npm')?.formatGlob ?? '**/*.js')
+  }
   hook = 'format:staged'
 }
 


plugins/lint-staged-npm/src/index.ts
```diff
@@ -4,14 +4,26 @@ import { getOptions } from '@dotcom-tool-kit/options'
...
   static description = 'format prettier'
 
-  key = getOptions('@dotcom-tool-kit/lint-staged-npm')?.testGlob ?? '**/*.js'
+  _key?: string
+  get key(): string {
+    if (!this._key) {
+      this._key = getOptions('@dotcom-tool-kit/lint-staged-npm')?.testGlob ?? '**/*.js'
+    }
+    return this._key
+  }
```
apaleslimghost, line 13: **suggestion**:

```suggestion
  get key(): string {
    return this._key ||= getOptions('@dotcom-tool-kit/lint-staged-npm')?.testGlob ?? '**/*.js'
  }
```

maybe?

apaleslimghost (approved):

plugins/lint-staged-npm/src/index.ts
ivomurrell: I've never seen that operator before!

apaleslimghost: i mean it's less than ideal because it's for booleans but `??=` doesn't exist so

ivomurrell: actually that's what I used in the end! https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment

apaleslimghost: WHATTTT

apaleslimghost: TIL

ivomurrell: Transpiled in node 12 but still

ivomurrell: beautiful
```js
return ((_a = this._key) !== null && _a !== void 0 ? _a : (this._key = (_c = (_b = (0, options_1.getOptions)('@dotcom-tool-kit/lint-staged-npm')) === null || _b === void 0 ? void 0 : _b.testGlob) !== null && _c !== void 0 ? _c : '**/*.js'));
```

apaleslimghost: 🤮

PR #159, opened 2022-01-07
CPP-774 Use release-please to manage package releases

ivomurrell:
This PR adds release-please integration to Tool Kit. This will allow us to make package versions independent of one another, whilst still enforcing compatibility via dependency ranges. It also removes `athloi` from the project, as we now manage the monorepo through a combination of npm workspaces and release-please, significantly improving install times. In order for release-please to be able to infer version bumps when doing releases, the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard is now enforced in the project.

The main drawback right now is the lack of proper support for prereleases in release-please. The issue is being tracked at https://github.com/googleapis/release-please/issues/1355, though we can create a fork with a fix if necessary in the future.


ivomurrell:
I said `--dry-run` damn

plugins/circleci/src/circleci-config.ts
```diff
@@ -70,11 +58,10 @@ export default abstract class CircleCiConfigHook extends Hook {
     // Automatically update tool kit orb if config is autogenerated
     const rawConfig = await this.getCircleConfigRaw()
     if (rawConfig && rawConfig.startsWith(automatedComment)) {
-      const targetTag = await this.getVersionTag()
       const currentOrb = config?.orbs['tool-kit']
       const currentTag = currentOrb && currentOrb.split('@')[1]
 
-      if (targetTag && targetTag !== developmentVersion && currentTag && targetTag !== currentTag) {
+      if (majorOrbVersion !== currentTag) {
```
apaleslimghost, line 64: **question**: do you have a process in mind for managing these versions? should that be documented?

**question**: this seems to be checking the major version against the entire version number in the `config.yml`, is that right?

readme.md
```diff
@@ -174,11 +174,23 @@ You can check if it's installed by running the help command:
...
+
+This means you should make an effort to think carefully about whether the changes you're making are a new feature or bug fix, and whether they contain any breaking changes. This might seem burdensome at first but it's good practice to make sure you can predict whether other teams' builds are going to break because of your code changes! If your commit will only affect a single package then please also include the name of the package (without the `@dotcom-tool-kit` namespace) in the scope of your commit message, as this makes it easier to see where changes are being made just by a quick glance at the git log. For example, a commit message for a new feature for the `circleci` plugin might look like:
+```
+feat(circleci): add support for nightly workflows
+```
+
```
apaleslimghost, line 193: **praise**: nice documentation

scripts/create-plugin.js
```diff
@@ -24,7 +24,7 @@ console.log('🔣 adding metadata to package.json')
 const pkg = JSON.parse(fs.readFileSync('package.json'))
 
 pkg.main = 'lib'
-pkg.version = '0.0.0-development'
+pkg.version = '0.1.0'
```
apaleslimghost, line 27: **question**: is the intention here for an in-development plugin to be created at 0.1, then release-please will bump that to 1.0? or will it do that without a breaking change?

apaleslimghost (approved):
i have a few questions, but this looks good if you can solve the bootstrapping peerdependencies issue :)

scripts/create-plugin.js
apaleslimghost: sounds good. it's a little counterintuitive to me that it needs a "breaking" change to go to 1.0 but i guess that's just a convention, and it won't happen often?

ivomurrell: I'd imagine in most cases we'd want to do a commit with a `release-as` footer when going to a 1.0 just so it's a bit clearer that it's going stable. To be honest, the alternative behaviour where breaking changes cause a minor bump for premajor versions just seems more intuitive to me, but it might be surprising for others.

apaleslimghost: i think we should do minor releases for 0.x breaking changes, and i like the `release-as` for 1.0 (as long as that's documented/well-known!)

ivomurrell: cool, I'll update the config and documentation then!

plugins/circleci/src/circleci-config.ts
ivomurrell: I've now added the orb to the release-please manifest so it can be released just like our node packages (its version is tracked with a `version.txt` file.) Is that what you had in mind?

Fixed the bug when checking the major version too, thank you for spotting that!

PR #163, opened 2022-01-26
Implement the Babel plugin

ivomurrell:


plugins/babel/src/run-babel.ts
```diff
@@ -0,0 +1,44 @@
...
+  // the outputted files.
+  // E.g., a glob of 'src/**/*.js'   = src/a/b.js -> lib/a/b.js
+  //       a glob of 'src/a/**/*.js' = src/a/b.js -> lib/b.js
+  const { base } = fg.generateTasks(fileGlob)[0]
+
+  const outputPath = options.outputPath ?? 'lib'
+  await fs.mkdir(outputPath, { recursive: true })
+
+  if (options.configFile) {
+    const { configFile } = options
```
leannecornish-ft, line 27: Question: will having this as an option allow users to pass in their own Babel options?

plugins/babel/src/run-babel.ts
```diff
@@ -0,0 +1,44 @@
...
+export async function runBabel(
+  logger: Logger,
+  options: BabelOptions,
+  transformOptions?: babel.TransformOptions
+): Promise<void> {
+  const fileGlob = options.files ?? 'src/**/*.js'
+  const files = await fg(fileGlob)
+  // Work out the root of the glob so we can strip this part of the path out of
+  // the outputted files.
+  // E.g., a glob of 'src/**/*.js'   = src/a/b.js -> lib/a/b.js
```
serena97, line 19: **praise**: helpful comment 👍 

**suggestion**: think it'll be useful to log out that we're running babel with the options here, so when people do npm run start and have babel configured into their run:local hook, they can see that babel is being run in their terminal

serena97 (approved):
a couple of small suggestions but should be ready to merge!

plugins/babel/src/run-babel.ts
ivomurrell: We expect users to set Babel options via a config file. This setting just (optionally) tells Babel where to find that config file: one thing Babel that can't read from the config file itself!

leannecornish-ft: Just wondering because in the ticket it says:
-do not allow passing in arbitrary Babel options to the tasks! should use the default Babel config files only (this is handled by @babel/core already)

ivomurrell: This option just tells `@babel/core` where to find those default config files

PR #164, opened 2022-01-28
CPP-771 Fix webpack failing if configPath option isn't set

ivomurrell:
If configPath wasn't set we were instead falling back to inserting an empty string into the list of arguments to pass to webpack. Unfortunately, unlike on a terminal where empty space would typically be stripped, the empty string is still interpreted as an argument in this case (similar to how the empty string would be an argument if you ran `webpack ''` instead of `webpack  `) and by default is considered a path to the webpack config file, which then isn't found. The webpack error isn't particularly helpful so it went unnoticed.

As a bonus, also pass the `--color` option to webpack to force ANSI colours even though we're piping to winston.


leannecornish-ft (approved):
🙌

PR #165, opened 2022-01-28
CPP-770 Fix issues with published backend-app package

ivomurrell:
- Corrected the `main` field to point to the `index.js` script
- Don't filter out `index.js` from files included in the published package by removing the `files` field
- Delete the unused `@dotcom-tool-kit/types` dependency

### DIFF
--- a/plugins/backend-app/package.json
+++ b/plugins/backend-app/package.json
@@ -2,12 +2,11 @@
   "name": "@dotcom-tool-kit/backend-app",
   "version": "0.0.0-development",
   "description": "",
-  "main": "lib",
+  "main": "index.js",
   "keywords": [],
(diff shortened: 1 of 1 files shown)


serena97 (approved):
thanks for spotting and fixing it! so sorry, I have forgotten to amend it after I used the create plugin script to generate the backend app

ivomurrell:
> while i've got you here: i'm surprised this package didn't break the integration tests, surely the incorrect `main` would have meant it couldn't be loaded?

This will be because the integration test checks the packages from the monorepo on the file system rather than pulling them from the npm registry. It would be nice to check both but not sure how we'd break the loop if a published package starts failing and preventing the CI from reaching the publish job for a patched version...

PR #166, opened 2022-01-28
persist workspace after build job

apaleslimghost:
`build` will probably have generated some files that we need to keep for future jobs


ivomurrell (approved):
Nevertheless, the build persisted

PR #168, opened 2022-02-07
cpp-781 Set vault env vars in heroku prod

serena97:
this is so that apps using toolkit, like next-health, can have their new prod AWS keys set through CI

### DIFF
--- a/plugins/heroku/src/setSlug.ts
+++ b/plugins/heroku/src/setSlug.ts
@@ -3,8 +3,9 @@ import { ToolKitError } from '@dotcom-tool-kit/error'
 import { readState } from '@dotcom-tool-kit/state'
 import { gtg } from './gtg'
 import type { Logger } from 'winston'
+import { setConfigVars } from './setConfigVars'
 
-function setSlug(logger: Logger, slug: string): Promise<void[]> {
+async function setSlug(logger: Logger, slug: string, systemCode?: string): Promise<void[]> {
   const state = readState(`production`)
 
   if (!state) {
@@ -14,6 +15,7 @@ function setSlug(logger: Logger, slug: string): Promise<void[]> {
   const appIds = state.appIds
 
   logger.info(`updating slug id ${slug} on production app ${appIds}`)
+
   const latestRelease = appIds.map((appId) =>
     heroku
       .post(`/apps/${appId}/releases`, {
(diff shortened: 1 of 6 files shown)


plugins/heroku/src/setSlug.ts
```diff
@@ -30,6 +32,10 @@ function setSlug(logger: Logger, slug: string): Promise<void[]> {
       })
       .then((response) => gtg(logger, response.app.name, 'production', false))
   )
+  
+  for(const id of appIds) {
+    await setConfigVars(logger, id, 'production', systemCode)
```
ivomurrell, line 37: **question**: why do we set the vault vars in the `setSlug` function? are the two actions related?

plugins/heroku/test/setSlug.test.ts
```diff
@@ -28,6 +35,13 @@ jest.mock('@dotcom-tool-kit/state', () => {
 })
 
 describe('setSlug', () => {
+  it('calls setConfigVars for each appIds', async () => {
+    mockHerokuPost.mockImplementationOnce(async () => Promise.resolve(goodHerokuResponse[0]))
+    mockHerokuPost.mockImplementationOnce(async () => Promise.resolve(goodHerokuResponse[1]))
+    await setSlug(logger, slugId)
+    expect(setConfigVars).toBeCalledTimes(2)
```
ivomurrell, line 42: **praise**: good that you're in the habit of writing tests for new code you add, I need to get better at that!

**suggestion**: would be good to check that the mock function is being called with the expected system code
```suggestion
    const systemCode = 'test-app'
    await setSlug(logger, slugId, systemCode)
    expect(setConfigVars).toHaveBeenNthCalledWith(1, logger, 'app-id-1', 'production', systemCode)
    expect(setConfigVars).toHaveBeenNthCalledWith(2, logger, 'app-id-2', 'production', systemCode)
```

ivomurrell (commented):
**question**: Is this the replacement for the `configure` task in `n-heroku-tools`? https://github.com/Financial-Times/n-heroku-tools/blob/main/tasks/configure.js Do we want to include any other functionality from that task, e.g., special handling of certain variable names?

Good spot that we weren't previously setting the environment variables in Tool Kit 👀

plugins/heroku/src/setSlug.ts
serena97: you're right, the naming isn't very helpful - I put it in setSlug because setSlug is only called in heroku/src/tasks/production.ts, and setSlug is the place where we get the appIds.

I'm thinking to rename 'setSlug' to 'promoteStagingToProduction'. what are your thoughts?

ivomurrell: I'd say it would be better to split it out into a different function and either call both this and `setSlug` from a `promoteStagingToProduction` function or just directly from the production task. I expect that we'll want to add more logic setting the environment variables anyway

PR #170, opened 2022-02-09
Only lint staged files in pre-commit hook

ivomurrell:
Should significantly speed up the time it takes to run the hook, which would previously often run for upwards of 10 seconds every time a `git commit` was invoked. We don't need to run a whole lint each time; during a commit we only care about verifying the staged files. Other files may have issues but that's okay as we aren't committing them right now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   ],
   "husky": {
     "hooks": {
-      "pre-commit": "npm run lint && lint-staged",
+      "pre-commit": "lint-staged",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -46,7 +46,7 @@
   ],
   "husky": {
     "hooks": {
-      "pre-commit": "npm run lint && lint-staged",
+      "pre-commit": "lint-staged",
```
apaleslimghost, line 49: shocked pikachu

PR #171, opened 2022-02-10
cpp-778 update config vars on review apps at first possible moment

leannecornish-ft:
We currently wait for the review-app to return a 'created' before applying config vars, which is a problem as the `aws_access_hashed_assets` is used during the `heroku-postbuild` script where the `release:remote` hook is called.

We can actually update the conf vars for all review-apps (see [api ref](https://devcenter.heroku.com/articles/platform-api-reference)) which is what n-gage is actually doing. Therefore, this PR is changing the `deploy:review` hook to apply the conf vars as as soon as we have a pipeline id, which _should_ mean they are there in time for the when the postbuild script is run.

### DIFF
--- a/plugins/heroku/src/setConfigVars.ts
+++ b/plugins/heroku/src/setConfigVars.ts
@@ -7,7 +7,8 @@ async function setConfigVars(
   logger: Logger,
   appIdName: string,
   environment: Environment,
-  systemCode?: string
+  systemCode?: string,
+  pipelineId?:string
 ): Promise<void> {
   try {
     logger.info(`setting config vars for ${appIdName}`)
@@ -22,7 +23,11 @@ async function setConfigVars(
       configVars.SYSTEM_CODE = systemCode
     }
 
-    await heroku.patch(`/apps/${appIdName}/config-vars`, { body: configVars })
+    const endpoint = appIdName === 'review-app' ? 
+      `/pipelines/${pipelineId}/stage/review/config-vars`:
+      `/apps/${appIdName}/config-vars`
+
+    await heroku.patch(endpoint, { body: configVars })
 
     logger.verbose('the following values have been set:', Object.keys(configVars).join(', '))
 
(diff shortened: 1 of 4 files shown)


plugins/heroku/src/setConfigVars.ts
```diff
@@ -22,7 +23,11 @@ async function setConfigVars(
       configVars.SYSTEM_CODE = systemCode
     }
 
-    await heroku.patch(`/apps/${appIdName}/config-vars`, { body: configVars })
+    const endpoint = appIdName === 'review-app' ? 
+      `/pipelines/${pipelineId}/stage/review/config-vars`:
+      `/apps/${appIdName}/config-vars`
```
apaleslimghost, line 28: **suggestion**: i think it would be cleaner to split this into two separate methods. that would make the logging below clearer as well.

apaleslimghost (approved):
great idea! one suggestion but looks good overall :)

plugins/heroku/src/setConfigVars.ts
leannecornish-ft: thanks - I've updated with what I think you mean! 😄

plugins/heroku/src/setConfigVars.ts
```diff
@@ -22,11 +23,16 @@ async function setConfigVars(
       configVars.SYSTEM_CODE = systemCode
     }
 
-    await heroku.patch(`/apps/${appIdName}/config-vars`, { body: configVars })
+    if (appIdName === 'review-app') {
+      await heroku.patch(`/pipelines/${pipelineId}/stage/review/config-vars`, { body: configVars })
+    } else {
+      await heroku.patch(`/apps/${appIdName}/config-vars`, { body: configVars })
+    }
```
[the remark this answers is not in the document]
leannecornish-ft: ooooh haha. It does. I'll do that 👍

plugins/heroku/src/setConfigVars.ts
```diff
@@ -25,15 +27,46 @@ async function setConfigVars(
...
+    if (err instanceof Error) {
+      error.details = err.message
+    }
+    throw error
+  }
+}
+
+async function setStageConfigVars(
+  logger: Logger,
+  stage: Stage,
```
apaleslimghost, line 44: **praise**: keeping it flexible like this is really nice (as well as using the union type)!

apaleslimghost (approved):
exactly what i had in mind, great work!

PR #173, opened 2022-02-17
CPP-732 Add npm publishing to Tool Kit

serena97:
- moved npm publish task to npm plugin 
- created circleci-npm plugin containing publish:tag hook that links to NpmPublish task
- wrote CIRCLE_TAG to state in circleci plugin
- implemented additional logging and error messages
- added tool-kit/publish job in orb 

[[circle ci link](https://app.circleci.com/pipelines/github/Financial-Times/n-exponea/106/workflows/afa53c04-c3b3-41ea-bc5e-a8448549a76b)] [[npm registry link](https://www.npmjs.com/package/@financial-times/n-exponea/v/0.0.0-beta.11)] n-exponea successful build with tool-kit/publish


plugins/circleci-npm/src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const tasks = [NpmPublish]
+
+class PublishHook extends CircleCiConfigHook {
+  job = 'tool-kit/publish'
+  jobOptions = {
+    requires: ['tool-kit/test'],
+    filters: {
+      branches: { ignore: '/.*/' },
+      tags: { only: `${semVerRegex}` }
```
apaleslimghost, line 12: **issue**: CircleCI will only run jobs on tags if they've got a `tags:` filter. because the other jobs in the workflow don't have this filter, they won't run at all. this hook (or something else?) will need to somehow add that filter to every other job in the config, e.g. [like in n-exponea](https://github.com/Financial-Times/n-exponea/blob/main/.circleci/config.yml#L25-L27)

plugins/npm/src/tasks/npm-publish.ts
```diff
@@ -0,0 +1,61 @@
...
+    this.logger.info(`tag ${tag} ready to be published`)
+
+    if (!process.env.NPM_AUTH_TOKEN) {
+        throw new ToolKitError('NPM_AUTH_TOKEN environment variable not found! Make sure your project is pulling in the env vars from /teams/next/circleci/component')
+    }
+
+    // overwrite version from the package.json with the version from e.g. the git tag
+    manifest.version = tag.replace(/^v/, '')
+
+    const tarball = await pack(packagePath)
```
apaleslimghost, line 50: **suggestion**: to help debug publishes it would be really useful if this task could print the list of files that's been packed here. don't worry about it if it gets too complicated though

plugins/circleci-npm/readme.md
```diff
@@ -0,0 +1,33 @@
+# `circleci-npm` Tool Kit plugin
+
+This plugin is for managing Tool Kit hooks that are run from npm scripts (such as `npm run test`).
```
apaleslimghost, line 3: **nitpick**: this description is copied from the `npm` plugin, but it's not accurate for this plugin

plugins/circleci-npm/src/index.ts
serena97: working on this one at the moment - from what I understand we should only add the tags filter to other jobs in the workflow if there is a hook included that uses the tags filter. so for apps that use circleci-heroku plugin eg. next-static, we shouldn't need to add the tags filter to all of its jobs as there's no jobs that would run with a specific tag, and I imagine we won't be thinking of adding a tags filter to any of its future jobs.

to ensure we only add the tags filter to all the other jobs in the workflow if tool-kit/publish is included, I think we could include a postinstall function and call it [here](https://github.com/Financial-Times/dotcom-tool-kit/blob/main/core/cli/src/install.ts#L14) to write the tags filter to config.yml if it sees that tool-kit/publish is in the config. 
another solution is to check if tool-kit/publish is already in circle config in [install() of circleci-config.ts](https://github.com/Financial-Times/dotcom-tool-kit/blob/main/plugins/circleci/src/circleci-config.ts#L189), and if so then write the tags filter to all other jobs. however I don't think it's the best solution because we can't guarantee the order of when the publish hook is written to circleci config. for example if we add circleci plugin in toolkit.yml before circleci-npm plugin then the hooks in circleci plugin get written to config.yml first. this solution is only viable if we introduce some sort of sorting mechanism of the plugin order but adding sort order seems more complicated than adding a post install step

plugins/npm/src/tasks/npm-publish.ts
```diff
@@ -0,0 +1,70 @@
...
+    if(!semVerRegex.test(tag)) {
+        throw new ToolKitError(`CIRCLE_TAG does not match regex ${semVerRegex}. Configure your release version to match the regex eg. v1.2.3-beta.8`)
+    }
+  }
+
+  async listPackedFiles(packagePath: string): Promise<void> {
+    this.logger.info('packing these files:')
+    readdirSync(packagePath).forEach(fileDir => {
+      this.logger.info(fileDir)
+    })
```
apaleslimghost, line 28: **issue/suggestion**: this is logging every file in the folder; but that's probably not the same as what's actually being published, it's not taking into account `.npmignore`, `.gitignore`, the `files` field in `package.json` etc. this could be pretty misleading! what you could probably do instead is read the actual list of files in the generated tarball (using https://github.com/npm/node-tar/ maybe?), so we know exactly what files are being published.

serena97:
n-exponea successful prerelease publish with tool-kit/publish: 
- [circleci](https://app.circleci.com/pipelines/github/Financial-Times/n-exponea/129/workflows/cb9ee112-5899-421c-935b-8902dc858dc9)
- [github release](https://github.com/Financial-Times/n-exponea/releases/tag/v0.0.0-beta.12)

plugins/npm/src/tasks/npm-publish.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    const tag = ci.tag
+
+    const npmTag = this.getNpmTag(tag)
+    
+    this.logger.info(`version ${tag} ready to be published with ${npmTag} tag`)
+
+    if (!process.env.NPM_AUTH_TOKEN) {
+        throw new ToolKitError('NPM_AUTH_TOKEN environment variable not found! Make sure your project is pulling in the env vars from /teams/next/circleci/component')
```
ivomurrell, line 62: **suggestion**: this should probably be updated to suggest checking the new GitHub contexts rather than Vault now

plugins/npm/src/tasks/npm-publish.ts
```diff
@@ -0,0 +1,82 @@
...
+  
+  getNpmTag(tag: string): TagType {
+    if(!tag) {
+        throw new ToolKitError('CIRCLE_TAG environment variable not found. Make sure you are running this on a release version!')
+    }
+    if(prereleaseRegex.test(tag)) {
+      return 'prerelease'
+    }
+    if(releaseRegex.test(tag)) { 
+      return 'latest'
```
ivomurrell, line 28: **thought**: I never realised we set these tags ourselves when publishing – cool!

ivomurrell (changes requested):
Sorry that there's still a lot of weird quirks related to the `circleci` plugin (that will hopefully go away after we allow state to be shared!) There are a few issues relating to that, and then a few nitpicks here and there, but this plugin is going to be super useful, thank you!

plugins/npm/src/tasks/npm-publish.ts
```diff
@@ -0,0 +1,82 @@
...
+    this.logger.info('preparing to publish your npm package....')
+
+    const packagePath = process.cwd()
+    const manifest = await pacote.manifest(packagePath)
+
+    const ci = readState('ci')
+    
+    if(!ci) {
+      throw new ToolKitError(
+        `Could not find state for ci, check that ${styles.hook('publish:tag')} ran successfully`
```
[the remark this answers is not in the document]
serena97: oops, good spot!

ivomurrell (approved):
This looks great now, thank you!

PR #175, opened 2022-02-22
add workflow for releasing prereleases

apaleslimghost:
for version tags with the `-beta.whatever` prerelease tags, exclude them from the main publish workflow, and publish them with a separate workflow that doesn't use the `latest` dist tag so they don't get picked up when doing `npm install dotcom-tool-kit@latest` or whatever

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,14 @@ references:
   filters_release_build: &filters_release_build
     tags:
       only:
-        - /^v\d+\.\d+\.\d+(?:-\w+\.\d+)?$/
+        - /^v\d+\.\d+\.\d+$/
+    branches:
+      ignore: /.*/
+  
+  filters_prerelease_build: &filters_release_build
+    tags:
+      only:
+        - /^v\d+\.\d+\.\d+(?:-\w+\.\d+)$/
     branches:
       ignore: /.*/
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -62,7 +62,14 @@ references:
   filters_release_build: &filters_release_build
     tags:
       only:
-        - /^v\d+\.\d+\.\d+(?:-\w+\.\d+)?$/
+        - /^v\d+\.\d+\.\d+$/
+    branches:
+      ignore: /.*/
+  
+  filters_prerelease_build: &filters_release_build
```
ivomurrell, line 69: **issue:** the CI is failing because the anchor name hasn't been updated after pasting

```suggestion
  filters_prerelease_build: &filters_prerelease_build
```

ivomurrell (changes requested):
I also didn't know this was possible to set, nor that it was an issue in the first place. Thanks for fixing this! One small copy+paste bug is breaking everything, though, as per

.circleci/config.yml
apaleslimghost: omg my god i was staring at the config for so long trying to work out what was wrong

ivomurrell (approved):

PR #176, opened 2022-02-25
CPP-782 Add nodemon plugin

ivomurrell:
This plugin provides a task that can be used in place of the Node plugin and will restart the running process if there are file changes in the directory or if the process exits unexpectedly. Will respect the configuration specified in a `nodemon.json` in `$CWD`, `$HOME`, or the location specified by the `configPath` option.

I figured there was big enough difference in code to justify splitting this out into a new module rather than just being an option you could toggle in a `.toolkitrc.yml`. Splitting the plugins also means that we do not force consumers to install the `nodemon` package when they just want to install the Node plugin.

### DIFF
--- a/core/sandbox/package.json
+++ b/core/sandbox/package.json
@@ -17,22 +17,21 @@
     "@dotcom-tool-kit/circleci-heroku": "file:../../plugins/circleci-heroku",
     "@dotcom-tool-kit/eslint": "file:../../plugins/eslint",
     "@dotcom-tool-kit/frontend-app": "file:../../plugins/frontend-app",
+    "@dotcom-tool-kit/jest": "file:../../plugins/jest",
     "@dotcom-tool-kit/lint-staged": "file:../../plugins/lint-staged",
     "@dotcom-tool-kit/lint-staged-npm": "file:../../plugins/lint-staged-npm",
     "@dotcom-tool-kit/mocha": "file:../../plugins/mocha",
-    "@dotcom-tool-kit/jest": "file:../../plugins/jest",
     "@dotcom-tool-kit/n-test": "file:../../plugins/n-test",
+    "@dotcom-tool-kit/next-router": "file:../../plugins/next-router",
+    "@dotcom-tool-kit/nodemon": "file:../../plugins/nodemon",
     "@dotcom-tool-kit/npm": "file:../../plugins/npm",
     "@dotcom-tool-kit/prettier": "file:../../plugins/prettier",
     "@dotcom-tool-kit/upload-assets-to-s3": "file:../../plugins/upload-assets-to-s3",
-    "dotcom-tool-kit": "file:../cli"
+    "dotcom-tool-kit": "file:../cli",
+    "nodemon": "^2.0.15"
   },
   "volta": {
     "extends": "../../package.json"
-  },
-  "dependencies": {
-    "@dotcom-tool-kit/next-router": "file:../../plugins/next-router",
-    "@dotcom-tool-kit/node": "file:../../plugins/node"
   },
   "husky": {
     "hooks": {
--- a/plugins/nodemon/src/tasks/nodemon.ts
+++ b/plugins/nodemon/src/tasks/nodemon.ts
@@ -15,7 +15,7 @@ export default class Nodemon extends Task<typeof NodemonSchema> {
   }
 
   async run(): Promise<void> {
-    const { entry } = this.options
+    const { entry, configPath } = this.options
     const vault = new VaultEnvVars(this.logger, {
       environment: 'development'
     })
@@ -42,7 +42,7 @@ export default class Nodemon extends Task<typeof NodemonSchema> {
       PORT: port.toString(),
       ...process.env
     }
-    nodemon({ script: entry, env, stdout: false })
+    nodemon({ script: entry, env, stdout: false, configFile: configPath })
     nodemon.on('readable', () => {
       // These fields aren't specified in the type declaration for some reason
       const { stdout, stderr } = (nodemon as unknown) as { stdout: Readable; stderr: Readable }
--- a/plugins/nodemon/src/tasks/nodemon.ts
+++ b/plugins/nodemon/src/tasks/nodemon.ts
@@ -1,0 +1,72 @@
+import { ToolKitError } from '@dotcom-tool-kit/error'
+import { hookFork, styles } from '@dotcom-tool-kit/logger'
+import { Task } from '@dotcom-tool-kit/types'
+import { NodemonOptions, NodemonSchema } from '@dotcom-tool-kit/types/lib/schema/nodemon'
+import { VaultEnvVars } from '@dotcom-tool-kit/vault'
+import getPort from 'get-port'
+import nodemon from 'nodemon'
+import { Readable } from 'stream'
+
+export default class Nodemon extends Task<typeof NodemonSchema> {
+  static description = ''
+
+  static defaultOptions: NodemonOptions = {
+    entry: './server/app.js'
+  }
+
+  async run(): Promise<void> {
+    const { entry } = this.options
+    const vault = new VaultEnvVars(this.logger, {
+      environment: 'development'
+    })
+
+    const vaultEnv = await vault.get()
+    const port =
+      Number(process.env.PORT) ||
+      (await getPort({
+        port: [3001, 3002, 3003]
+      }))
+
+    if (!entry) {
+      const error = new ToolKitError(
+        `the ${styles.task('Nodemon')} task requires an ${styles.option('entry')} option`
+      )
+      error.details = `this is the entrypoint for your app, e.g. ${styles.filepath('server/app.js')}`
+      throw error
+    }
+
+    this.logger.verbose('starting the child nodemon process...')
+
+    const env = {
+      ...vaultEnv,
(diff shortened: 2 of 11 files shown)


plugins/nodemon/src/tasks/nodemon.ts
```diff
@@ -0,0 +1,72 @@
...
+    const vaultEnv = await vault.get()
+    const port =
+      Number(process.env.PORT) ||
+      (await getPort({
+        port: [3001, 3002, 3003]
+      }))
+
+    if (!entry) {
+      const error = new ToolKitError(
+        `the ${styles.task('Node')} task requires an ${styles.option('entry')} option`
```
rowanmanning, line 32: **suggestion:** I suspect this may be a copy/paste error. Should this be `Nodemon`?

```suggestion
        `the ${styles.task('Nodemon')} task requires an ${styles.option('entry')} option`
```

I don't know how common logging the name is in other plugins, but potentially using `this.contructor.name` would avoid copy/paste in future and also allow extension without having to override methods. _However_ maybe extending tasks is frowned upon in which case ignore!

plugins/nodemon/src/tasks/nodemon.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+    nodemon({ script: entry, env, stdout: false })
+    nodemon.on('readable', () => {
+      // These fields aren't specified in the type declaration for some reason
+      const { stdout, stderr } = (nodemon as unknown) as { stdout: Readable; stderr: Readable }
+      hookFork(this.logger, entry, { stdout, stderr })
+    })
+    const nodemonLogger = this.logger.child({ process: 'nodemon' })
+    nodemon.on('log', (msg) => {
+      function nodemonToWinstonLogLevel(level: string): string {
```
rowanmanning, line 53: **nitpick:** I don't think this will have a particularly negative impact if it's not addressed, but this function is redefined each time `nodemon` logs. It's maybe a bit of a micro-optimisation to move it out of here though so happy if you want to ignore.

rowanmanning (commented):
Very excited for this to release, thanks!

I haven't done an in-depth review but, while I was investigating how you got this working, I spotted a few potential issues/improvements. Please don't consider the nitpick to be blocking, as tbh it could come down to tool kit code style which I'm not familiar with.

plugins/nodemon/src/tasks/nodemon.ts
```diff
@@ -0,0 +1,72 @@
...
+import getPort from 'get-port'
+import nodemon from 'nodemon'
+import { Readable } from 'stream'
+
+export default class Nodemon extends Task<typeof NodemonSchema> {
+  static description = ''
+
+  static defaultOptions: NodemonOptions = {
+    entry: './server/app.js'
+  }
```
rowanmanning, line 15: **question:** something else I suddenly thought of is that [in dotcom-debug-user](https://github.com/Financial-Times/dotcom-debug-user/blob/main/.toolkitrc.yml#L34), we're having to specify `jsx` as a file extension quite often. Is adding in `jsx` by default something that Platforms would consider? I don't feel like it would cause issues for projects which don't use JSX for templating.

[Nodemon makes this configurable with the `ext` option](https://github.com/remy/nodemon/#specifying-extension-watch-list), so I _think_ this could be adjusted to something like this (which adds `jsx` to Nodemon's defaults and ditches some of the extensions we don't use like CoffeeScript):

```suggestion
  static defaultOptions: NodemonOptions = {
    entry: './server/app.js',
    ext: ['cjs', 'js', 'json', 'jsx', 'mjs']
  }
```

(would need to update the types as well I think)

plugins/nodemon/src/tasks/nodemon.ts
ivomurrell: I have a habit of embedding throwaway functions like this right next to the code that uses it so that you don't have to jump around too much when reading the code. It's the kind of thing that I'd hope V8 will optimise away at runtime though, especially when it's not even using any closure state! Maybe that's too optimistic haha I have no idea about the engine's internals

plugins/nodemon/src/tasks/nodemon.ts
ivomurrell: Nice catch! I hope that we can make this error handling generic at some point; many tasks have to handle mandatory config options. I'm going to stick with hardcoding the name for now so that the message is a bit more readable and we don't want to actually have to be copy + pasting this code everywhere anyway!

plugins/nodemon/src/tasks/nodemon.ts
ivomurrell: This is probably something that should be done with a [`nodemon.json` config file in your project](https://github.com/remy/nodemon#config-files) rather than having Toolkit handle it specifically. I don't think we should be overriding any configuration defaults within Tool Kit if we can help it, although `.jsx` support would be a pretty common use case. What do you think, @apaleslimghost?

apaleslimghost: yep, the general philosophy for Tool Kit is not to provide another method of configuration if the underlying tool has its own way of doing that (e.g. we don't allow specifying Webpack config options in `.toolkitrc.yml`). honestly? i'd open a PR on Nodemon to add `.jsx` to the default extensions 😅

core/sandbox/package.json
```diff
@@ -32,7 +32,8 @@
   },
   "dependencies": {
     "@dotcom-tool-kit/next-router": "file:../../plugins/next-router",
-    "@dotcom-tool-kit/node": "file:../../plugins/node"
+    "@dotcom-tool-kit/nodemon": "file:../../plugins/nodemon",
+    "nodemon": "^2.0.15"
```
apaleslimghost, line 36: **nitpick (super inconsequential)**: these should all be `devDependencies`. doesn't matter at all but might be slightly confusing if anybody uses the sandbox as a reference

plugins/nodemon/src/tasks/nodemon.ts
rowanmanning: I was tempted to go and do this, but then [this discussion](https://github.com/remy/nodemon/issues/1954) about adding `cjs` didn't really go anywhere and the [matching PR](https://github.com/remy/nodemon/pull/1957) hasn't been touched since November so I might just leave it. Fine to continue configuring ourselves as we have to do this for other plugins anyway 👍

core/sandbox/package.json
```diff
@@ -17,23 +17,22 @@
...
-    "@dotcom-tool-kit/jest": "file:../../plugins/jest",
     "@dotcom-tool-kit/n-test": "file:../../plugins/n-test",
+    "@dotcom-tool-kit/next-router": "file:../../plugins/next-router",
+    "@dotcom-tool-kit/nodemon": "file:../../plugins/nodemon",
     "@dotcom-tool-kit/npm": "file:../../plugins/npm",
     "@dotcom-tool-kit/prettier": "file:../../plugins/prettier",
     "@dotcom-tool-kit/upload-assets-to-s3": "file:../../plugins/upload-assets-to-s3",
-    "dotcom-tool-kit": "file:../cli"
+    "dotcom-tool-kit": "file:../cli",
+    "nodemon": "^2.0.15"
```
serena97, line 31: nitpick: nodemon is a peerdep in the plugin, so we can remove this

plugins/nodemon/src/tasks/nodemon.ts
```diff
@@ -42,7 +42,7 @@ export default class Nodemon extends Task<typeof NodemonSchema> {
       PORT: port.toString(),
       ...process.env
     }
-    nodemon({ script: entry, env, stdout: false })
+    nodemon({ script: entry, env, stdout: false, configFile: configPath })
```
serena97, line 45: nitpick: configPath can potentially be undefined here because we don't have a defaultOption for it. If this is causing nodemon to log undefined I think we should handle the case when configPath is undefined

serena97 (approved):
it looks good to @leannecornish-ft and I, we have a couple of non blocking nitpicks but this is good to merge

core/sandbox/package.json
ivomurrell: Even though peer dependencies will be installed automatically I think it's still a good idea to manually specify the version in the end product if we are explicitly using it – like we are in our `package.json`'s `"scripts"` field – so that the major version won't suddenly change if the peer dependency's version range is widened.

plugins/nodemon/src/tasks/nodemon.ts
ivomurrell: Where is `configPath` (or `configFile`) logged?

The default option for `configFile` is `undefined` as we want it to look for the config file in the default search paths (`$PWD` and `$HOME`) if `configPath` isn't set in Tool Kit's config.

ivomurrell: This is more like calling `nodemon --config example.json server/app.js`, which does not affect the `node` invocation that is logged. Setting the `configFile` argument to `undefined` is equivalent to not setting it at all, causing `nodemon` to look up the config file in its default search paths instead.

serena97: ah fair enough, if setting the configFile argument to undefined in the api won't reflect as undefined in the logs then that's fine

PR #183, opened 2022-03-16
create build-test job to stop workspace persistence

leannecornish-ft:
This is a workaround to avoid the toolkit state file being overwritten. Apps can use this job in their .circleci/config.yml job rather than build and test, as used in next-static [here](https://github.com/Financial-Times/next-static/blob/4cdf82ce537f30eaa0d646d0e0fadd236acf4b5c/.circleci/config.yml#L51-L62)

### DIFF
--- a/orb/src/jobs/build-test.yml
+++ b/orb/src/jobs/build-test.yml
@@ -1,0 +1,26 @@
+# temp workaround for heroku-staging's .toolkitstate.json being overwritten by the build job
+parameters:
+  node-version:
+    default: '12.22'
+    type: string
+
+executor:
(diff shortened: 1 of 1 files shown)


apaleslimghost (approved):
this is fine, but:

**question**: does this need to go into the orb? could we temporarily define it in the apps if it's just a workaround?

**question**: what's the long-term fix?

leannecornish-ft:
Yes, I guess people could add it but I think it will be easier for them to add 3 lines of code (the rest needs to be done in either case)?

The long term fix is to separate out the toolkitstate.json

PR #190, opened 2022-03-25
CPP-753 add documentation on how to write a custom plugin

apaleslimghost:


docs/custom-plugins.md
```diff
@@ -0,0 +1,143 @@
...
+  'build:remote: Rollup
+```
+
+You should install `@dotcom-tool-kit/types` and the tooling you're implementing as `devDependencies` of your repo (e.g. `npm install --save-dev @dotcom-tool-kit/types rollup`).
+
+## Implementing a new hook
+
+A Hook defines an abstract label to run tasks with, as well as managing where in other project configuration it's run from. For example, the built-in `circleci` plugin defines a `build:ci` hook, which lets tasks like Rollup run in CI, and it specifies that `build:ci` should be run by a CircleCI job and automatically manages the configuration in `.circleci/config.yml` to run that job.
+
+This abstraction lets us write different plugins for defining tasks to be run and where they should be run from, while maintaining the link between them. We've already seen that `build:ci` could be running Rollup, or Webpack, or any other task; in addition, `build:ci` itself could be defined by a different plugin, such as a Github Actions plugin, that would automatically manage configuration in `.github/workflows`.
```
ivomurrell, line 81: **suggestion**: I found the first sentence a little hard to parse
```suggestion
This abstraction lets us write different plugins for defining tasks to be run, separate from the plugins defining where they should be run from, whilst maintaining the link between them. We've already seen that `build:ci` could be running Rollup, or Webpack, or any other task; in addition, `build:ci` itself could be defined by a different plugin, such as a Github Actions plugin, that would automatically manage configuration in `.github/workflows`.
```

docs/custom-plugins.md
```diff
@@ -0,0 +1,143 @@
...
+
+If you find yourself asking a question like "how do I run a Tool Kit task from a different npm script", **you should implement a hook** to allow Tool Kit to automatically manage that configuration for any new repos using your plugin, rather than expecting new users to add that configuration themselves when installing the plugin.
+
+Hooks have a loose naming convention of `category:environment`. This is only meant for humans to be able to intuitively understand which hooks are related; it's not required by the Tool Kit core itself.
+
+Let's say you want to run some task on the npm `prepare` script (which automatically runs after `npm install` and before `npm publish`). We'll call that hook `prepare:local`. Create a subclass of the `Hook` class from `@dotcom-tool-kit/types`, implement the `check` and `install` methods, and export a `hooks` object to map it to the name we're giving it. Your `index.js` might include:
```
leannecornish-ft, line 89: I think it may be useful here to explain that a hook would be in a subdirectory in `toolkit` with an appropriate name, e.g. prepare(?). Also that the index file can contain the hook alone, or be in conjunction with a task. Or actually maybe do we want it do be only in conjunction with a task?

leannecornish-ft (commented):
I'm too late 😂

leannecornish-ft:
Over all it looks really great though 👏

PR #191, opened 2022-03-25
cpp-830 split state out into different files

leannecornish-ft:
Currently we have one state file in toolkit - .toolkitstate.json in the project's root. This is causing a problem in circle ci, where multiple concurrent jobs are writing to the state file and persisting that to the next job. Where a job attaches both versions (via workspaces), the last copy overwrites the any others, thereby losing the state information.

This PR splits the state into different stages, thereby allowing for writes from multiple concurrent jobs/workspaces.

### DIFF
--- a/lib/state/src/index.ts
+++ b/lib/state/src/index.ts
@@ -1,6 +1,8 @@
+import path from 'path'
 import * as fs from 'fs'
+
 const target = process.env.INIT_CWD || process.cwd()
-const stateDir = target ? `${target}/.toolkitstate` : '/.toolkitstate'
+const stateDir = target ? path.join(target, '/.toolkitstate') : '.toolkitstate'
 
 interface CIState {
   repo: string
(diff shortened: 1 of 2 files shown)


lib/state/src/index.ts
```diff
@@ -37,30 +37,33 @@ export interface State {
...
-    }
-    fs.writeFileSync(stateFile, JSON.stringify(readStateContent, null, 2))
+  const file = `${stateDir}/${stage}.json`
+  let readStateContent
+  if (fs.existsSync(stateDir)) {
+    if (fs.existsSync(file)) {
+      readStateContent = JSON.parse(fs.readFileSync(file, { encoding: 'utf-8' }))
+      for (const [key, val] of Object.entries(value)) {
+        readStateContent[key] = val
+      }
```
apaleslimghost, line 62: **suggestion**: this could be done with `Object.assign` now:

```suggestion
      readStateContent = Object.assign(
        JSON.parse(fs.readFileSync(file, { encoding: 'utf-8' })),
        value
      )
```

lib/state/src/index.ts
```diff
@@ -1,6 +1,6 @@
 import * as fs from 'fs'
 const target = process.env.INIT_CWD || process.cwd()
-const stateFile = target ? `${target}/.toolkitstate.json` : '.toolkitstate.json'
+const stateDir = target ? `${target}/.toolkitstate` : '/.toolkitstate'
```
apaleslimghost, line 3: **suggestion**: this (and constructing the stage `.json` paths below) should use `path.join` instead of a template string, currently this probably won't work on Windows

**issue**: i don't think the leading `/` on the fallback here is correct, that would try to create the directory at the root of the filesystem (which will almost certainly fail)

lib/state/src/index.ts
```diff
@@ -1,6 +1,8 @@
+import path from 'path'
 import * as fs from 'fs'
+
 const target = process.env.INIT_CWD || process.cwd()
-const stateFile = target ? `${target}/.toolkitstate.json` : '.toolkitstate.json'
+const stateDir = target ? path.join(target, '/.toolkitstate') : '.toolkitstate'
```
apaleslimghost, line 5: **nitpick**: you don't need the `/` in a `path.join`:

```suggestion
const stateDir = target ? path.join(target, '.toolkitstate') : '.toolkitstate'
```

apaleslimghost (approved):
one tiny nitpick but this looks great!

PR #200, opened 2022-03-31
CPP-846 Allow Node version to be configured in CircleCI hook

ivomurrell:
Now that we've migrated to newer Node versions we need to update CircleCI configs to use the correct Node version too. But seeing as both Node 14 and Node 16 should be supported we can't just hardcode one version (though that might cause upgrade issues down the road anyway!) So you can now set the Node version you want in the Tool Kit config, with the default being Node 16.


apaleslimghost (approved):
this is great! but,

**suggestion**: could this use `engines` or the volta config? just to avoid one more place people have to configure this?

ivomurrell:
I think that latter idea could work! Something to explore at least.

Also just realised this doesn't support the 14/16 matrices. I guess we'll just have to ask people to maintain those manually for now if they need them...?

PR #202, opened 2022-04-14
remove babel plugin from the backend-app preset

apaleslimghost:
it won't necessarily be needed, and we don't have a good workflow in place for e.g. running it before starting the server or in watch mode, so remove it and let people configure it themselves


apaleslimghost:
it's potentially something we'll need to improve, but i think we'd need to work with people to understand the use case?

PR #213, opened 2022-04-22
Don't include Babel in frontend-app config

ivomurrell:
We removed Babel from the `backend-app` plugin recently, but we defined a conflict resolution between it and the Webpack plugin for the `frontend-app` plugin which is now causing an error seeing as the Babel hooks are not present in the plugin anymore.


apaleslimghost (approved):
oops, good spot

PR #214, opened 2022-04-22
Bump Tool Kit dependencies in the sandbox to 2.0

ivomurrell:


serena97 (approved):
nice spot. do we need to add core/sandbox to release-please-manifest.json and release-please-config.json too?

core/sandbox/.circleci/config.yml
```diff
@@ -90,6 +90,16 @@ workflows:
               ignore: /.*/
             tags:
               only: /^v\d+\.\d+\.\d+(-.+)?/
+      - tool-kit/publish:
```
serena97, line 93: 👀 why is this part added?

ivomurrell: It's a newer hook that's installed now, but this is just a test project so this config will never run and won't publish anything! 😁

ivomurrell:
No, we never release or publish the sandbox – it's just a test project

serena97:
I agree we shouldn't release or publish the sandbox, the reason I brought it up is because it would be cumbersome to update sandbox manually whenever we make a new release which happens quite often. perhaps we can automate this, by adding to the config.yml a script that runs after the release-please command?

ivomurrell:
I think the current behaviour is preferred: we only need to do a manual update when we release a new major version of one of its dependencies and very often those will have breaking changes which might mean we will have to change the code in the sandbox too. So we can't just automate the releases, and it's acceptable for the sandbox to lag behind the other packages and only be updated when someone is actually using it imo.

PR #215, opened 2022-04-22
Replace circleci job if needs updating, avoiding duplication

serena97:
pr [#214](https://github.com/Financial-Times/dotcom-tool-kit/pull/214) reveals that there is an issue where the publish job is added into the config.yml with the `npx dotcom-tool-kit --install`, when there is already an existing publish job. when we check for duplication in install() it does not account for the recently added node-version [property](https://github.com/Financial-Times/dotcom-tool-kit/blob/main/plugins/circleci/src/circleci-config.ts#L169), which causes the publish job to be added again. instead, this PR replaces the original circleci job with the new job if it needs updating


plugins/circleci/src/circleci-config.ts
```diff
@@ -164,17 +171,23 @@ export default abstract class CircleCiConfigHook extends Hook {
...
-      [this.job]: {
-        'node-version': getOptions('@dotcom-tool-kit/circleci')?.nodeVersion ?? '16.14-browsers',
-        ...this.jobOptions
-      }
+    const computedJob = this.getComputedJob()
+
+    const hasJob = (jobs: NonNullable<Workflow['jobs']>): boolean => {
+      return !jobs.some((candidateJob, index) => { 
+                if(Object.keys(candidateJob)[0] === this.job && !isEqual(candidateJob, computedJob)) {
+                  jobs[index] = computedJob; return true
```
ivomurrell, line 179: **todo**: would you be able to run this code through prettier so that it's a bit more readable?

ivomurrell, line 179: **nitpick**: I find it a bit surprising that a function called `hasJob` has side effects, could we move the assignment out of this function?

plugins/circleci/src/circleci-config.ts
```diff
@@ -186,7 +199,7 @@ export default abstract class CircleCiConfigHook extends Hook {
       nightlyJobs.push(nightlyJob)
     }
 
-    const serialised = automatedComment + yaml.dump(config)
+    const serialised = automatedComment + yaml.dump(config, {noRefs: true})
```
ivomurrell, line 202: **praise**: good idea!

ivomurrell (approved):
This looks good, just got a few minor comments!

joelcarr:
There have been breaking changes to how the CircleCI plugin works, this PR is no longer applicable.

PR #221, opened 2022-04-27
CPP-783 add webpack watch mode

apaleslimghost:


core/sandbox/.toolkitstate/local.json
```diff
@@ -0,0 +1,3 @@
+{
+  "port": 3001
```
ivomurrell, line 2: **issue**: I don't think this file should be tracked by git; this file is updated by the Node plugin on each run with whatever port it's chosen

plugins/frontend-app/.toolkitrc.yml
```diff
@@ -1,3 +1,8 @@
 plugins:
   - '@dotcom-tool-kit/webpack'
   - '@dotcom-tool-kit/backend-app'
+
+hooks:
+  'run:local':
+    - Node
+    - WebpackWatch
```
ivomurrell, line 8: **thought**: don't we want the package to have been built (once) with webpack before we run it with Node?

plugins/webpack/src/run-webpack.ts
```diff
@@ -5,12 +5,23 @@ import { hookFork, waitOnExit } from '@dotcom-tool-kit/logger'
 
 const webpackCLIPath = require.resolve('webpack-cli/bin/cli')
 
-export default function runWebpack(logger: Logger, options: WebpackOptions): Promise<void> {
+export interface RunWebpackOptions {
+  mode: 'production' | 'development',
```
ivomurrell, line 9: **praise**: this is much cleaner than having it statically set in the config

ivomurrell (commented):
This looks good!

One extra **suggestion**: could we set the `entry` field in the sandbox's `webpack.config.js` to `./server/app.js` so that it doesn't throw an error about not finding a `src` directory?

plugins/frontend-app/.toolkitrc.yml
apaleslimghost: hmm, yes, so; the webpack-built files aren't necessary for the server to run, they're just served alongside the node server usually. except, they actually are required in most cases, since the Page Kit asset loader tries to load the manifest generated by webpack on initialisation. i guess there's two potential solutions:

- add a `WebpackDevelopment` task here that runs before `Node`
- have `dotcom-server-asset-loader` lazily load the manifest

the second one seems hard so

serena97 (approved):
looks good to me! one small suggestion - it would be good to add some documentation in how to use this feature in webpack

serena97 (approved):

PR #222, opened 2022-04-27
CPP-878 detect conflicts between cousin plugins

apaleslimghost:
currently, we only detect conflicts between sibling plugin hooks, e.g.:

```
├ webpack
╰ babel
```

would correctly raise a conflict, because both are assigning the `build:*` hooks by default.

if, however, the packages are not direct siblings but cousins:

```
├ frontend-app
│ ╰ webpack
╰ babel
```

that conflict wouldn't be noticed, and Tool Kit would silently choose whichever plugin was loaded last to run on that hook without requiring conflict resolution.

this PR fixes that by updating the logic to only allow a plugin to override hooks/options set by their descendants (which was always the intent).

this is a breaking change, since it could mean that apps that work completely fine now might have silently ignored conflicts that this change would error on.

### DIFF
--- a/core/cli/src/plugin.ts
+++ b/core/cli/src/plugin.ts
@@ -11,6 +11,16 @@ import { ToolKitError } from '@dotcom-tool-kit/error'
 import { styles } from '@dotcom-tool-kit/logger'
 import { Hook, instantiatePlugin, Plugin, TaskClass } from '@dotcom-tool-kit/types'
 
+function isDescendent(possibleAncestor: Plugin, possibleDescendent: Plugin): boolean {
+  if (!possibleDescendent.parent) {
+    return false
+  } else if (possibleDescendent.parent === possibleAncestor) {
+    return true
+  } else {
+    return isDescendent(possibleAncestor, possibleDescendent.parent)
+  }
+}
+
 export async function loadPluginConfig(logger: Logger, plugin: Plugin, config: Config): Promise<Config> {
   const { plugins = [], hooks = {}, options = {} } = await loadToolKitRC(plugin.root)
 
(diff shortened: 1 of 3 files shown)


apaleslimghost:
okay so i guess in that example `webpack` is a nibling of `babel`. but really a nibling is a zeroth cousin once removed

ivomurrell:
> this is a breaking change, since it could mean that apps that work completely fine now might have silently ignored conflicts that this change would error on.

I disagree that this warrants a major version: this is fixing the behaviour so that it is working how it was always documented and intended, and I don't think anybody was relying on this behaviour. I think [this](https://semver.org/#what-if-i-inadvertently-alter-the-public-api-in-a-way-that-is-not-compliant-with-the-version-number-change-ie-the-code-incorrectly-introduces-a-major-breaking-change-in-a-patch-release) section of the semver spec is instructive:
> Use your best judgment. If you have a huge audience that will be drastically impacted by changing the behavior back to what the public API intended, then it may be best to perform a major version release, even though the fix could strictly be considered a patch release. Remember, Semantic Versioning is all about conveying meaning by how the version number changes. If these changes are important to your users, use the version number to inform them.

apaleslimghost:
it's not necessarily that users would have been relying on this behaviour, but that they'd expect a currently-working `.toolkitrc.yml` to keep working with a non-major update. but, releasing this as a patch version makes my life much easier, so,

core/cli/src/plugin.ts
```diff
@@ -11,6 +11,16 @@ import { ToolKitError } from '@dotcom-tool-kit/error'
 import { styles } from '@dotcom-tool-kit/logger'
 import { Hook, instantiatePlugin, Plugin, TaskClass } from '@dotcom-tool-kit/types'
 
+function isDescendent(possibleAncestor: Plugin, possibleDescendent: Plugin): boolean {
```
ivomurrell, line 14: **praise**: wow I did not know that descendent was a word

core/cli/test/index.test.ts
```diff
@@ -72,6 +72,21 @@ describe('cli', () => {
     expect(config).toHaveProperty('hookTasks.build:local.conflicting')
   })
 
+  it('should indicate when there are conflicts between plugins that are cousins in the tree', async () => {
```
ivomurrell, line 75: **thought**: maybe we should have another test to make sure that a plugin that is present in multiple ancestors doesn't cause a conflict?

e.g., that
```
├ frontend-app
│ ╰ webpack
╰ webpack
```
is okay, to give a contrived example.

ivomurrell (approved):
🤷🏻 for some reason I've always found this merging code really hard to grok, but I've tested it out and it seems to all work as you'd expect!

apaleslimghost:
> for some reason I've always found this merging code really hard to grok

i really regret using `_.mergeWith` and not just a for-loop

core/cli/test/index.test.ts
apaleslimghost: because of https://github.com/Financial-Times/dotcom-tool-kit/blob/main/core/cli/src/plugin.ts#L115-L118 there's no way it could conflict, but a test for that wouldn't hurt

PR #224, opened 2022-04-28
CPP-791 secret squirrel

apaleslimghost:
original implementation by @joelcarr [in next-newsletters-page](https://github.com/Financial-Times/next-newsletters-page/blob/main/toolkit/plugins/secret-squirrel/index.js) (thanks!)


plugins/backend-app/package.json
```diff
@@ -8,8 +8,9 @@
   "license": "ISC",
   "dependencies": {
     "@dotcom-tool-kit/circleci-heroku": "^2.0.0",
+    "@dotcom-tool-kit/node": "^2.0.0",
     "@dotcom-tool-kit/npm": "^2.0.0",
-    "@dotcom-tool-kit/node": "^2.0.0"
+    "@dotcom-tool-kit/secret-squirrel": "file:../secret-squirrel"
```
apaleslimghost, line 13: @ivomurrell i will need your input on how to release this cos i have no idea

plugins/backend-app/package-lock.json
```diff
@@ -0,0 +1,5754 @@
+{
```
ivomurrell, line 1: **issue**: hmm we shouldn't be making lock files for packages: they should all be managed from the root workspace. not sure why these have been generated by npm!?

plugins/secret-squirrel/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "directory": "plugins/secret-squirrel"
+  },
+  "bugs": "https://github.com/financial-times/dotcom-tool-kit/issues",
+  "homepage": "https://github.com/financial-times/dotcom-tool-kit/tree/main/plugins/secret-squirrel",
+  "files": [
+    "/lib",
+    ".toolkitrc.yml"
+  ],
+  "peerDependencies": {
+    "@financial-times/secret-squirrel": "^2.19.0",
```
ivomurrell, line 24: **suggestion**: we should make peer dependency ranges as wide as possible: https://nodejs.org/en/blog/npm/peer-dependencies/#using-peer-dependencies
```suggestion
    "@financial-times/secret-squirrel": "2.x",
```

plugins/secret-squirrel/readme.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+And add it to your repo's `.toolkitrc.yml`:
+
+```yml
+plugins:
+  - '@dotcom-tool-kit/secret-squirrel'
+```
+
+You will need a plugin that provides a hook to run the `SecretSquirrel` task. By default, the task runs on the `git:precommit` hook. One plugin that provides that hook is [`husky-npm`](../husky-npm); installing that alongside this plugin will enforce Secret Squirrel running on before every commit.
```
ivomurrell, line 20: **praise**: good to see docs being added at the same time a plugin is made, and with genuinely useful advice too!

plugins/backend-app/package.json
ivomurrell: the best I can come up with is to add the plugin to `backend-app` in a separate PR, sorry. though I'd also suggest that maybe we should be waiting for the `secret-squirrel` plugin to hit 1.0 before adding it to `backend-app` so that it doesn't cause any breaking changes here?

plugins/secret-squirrel/src/tasks/secret-squirrel.ts
```diff
@@ -0,0 +1,17 @@
+import { Task } from '@dotcom-tool-kit/types'
+import { fork } from 'child_process'
+import { hookFork, waitOnExit } from '@dotcom-tool-kit/logger'
+
+const secretSquirrelPath = require.resolve('.bin/secret-squirrel')
+
+export default class SecretSquirrel extends Task {
+  static description = 'run Secret Squirrel to check for secrets in your repo'
```
ivomurrell, line 8: **praise**: not just an empty string in the `description` field!
**thought**: filling these in should probably be part of the documentation work?

ivomurrell (commented):
It's good how nicely this plugin fits into the ecosystem!

serena97 (approved):
looks good to me! from what I understand, in terms of releasing I think once this is merged in, release please in the CI would open a new pr with the new secret-squirrel release, and since we aren't looking to add secret squirrel to the backend-app until v1.0.0 then no further action needs to be taken

PR #228, opened 2022-05-06
Fix tasks are not valid error

serena97:
The tasks are not valid error was introduced in commit [463124a](https://github.com/Financial-Times/dotcom-tool-kit/commit/463124add8fd89dea013dfa892a5ebebf8b5537c), where version ranges caused toolkit core/create to import installHooks from .npm node_modules instead of the app level node_modules. This caused the Task function location to be `.npm/_npx/4fc902e927bad53b/node_modules/@dotcom-tool-kit/types/lib/index.js:9` and the rawPlugin tasks prototype location to be `empty_app/node_modules/@dotcom-tool-kit/types/lib/index.js:9` thus making `instantiatePlugin` fail at the `task.prototype instanceof Task` typecheck, throwing 'tasks are not valid'. We fix this by making sure we import installHooks at the app level's node_modules.

### DIFF
--- a/core/create/src/index.ts
+++ b/core/create/src/index.ts
@@ -17,6 +17,7 @@ import prompt from 'prompts'
 import { promisify } from 'util'
 import { Logger } from './logger'
 import importFrom from 'import-from'
+import installHooksType from 'dotcom-tool-kit/lib/install'
 
 const exec = promisify(_exec)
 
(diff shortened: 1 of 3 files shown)


core/create/src/index.ts
```diff
@@ -139,7 +139,9 @@ async function executeMigration(deleteConfig: boolean) {
 
   packageJson.writeChanges()
 
-  const installPromise = logger.logPromise(exec('npm install'), 'installing dependencies')
+  await logger.logPromise(exec('npm install'), 'installing dependencies')
+  // we need to import installHooks from the app itself instead of npx or else loadPlugin will load rawPlugin from npx and Task will be loaded from the app, leading to task.prototype failing the instanceof Task check
```
apaleslimghost, line 143: **praise**: great explanation, really clear

core/create/src/index.ts
```diff
@@ -139,7 +139,9 @@ async function executeMigration(deleteConfig: boolean) {
...
+  const installHooks: any = ((await importFrom(process.cwd(), 'dotcom-tool-kit/lib/install')) as any).default
```
apaleslimghost, line 144: **question**: is is possible to cast this to the actual type of `installHooks`? even if we're doing `as unknown as typeof installHooks` i think that would be better?

**nitpick**: `importFrom` is synchronous so the `await` is unnecessary (same below)

apaleslimghost (approved):
nice! couple of minor comments but looks good

PR #229, opened 2022-05-06
Bump create version

serena97:
- the current release version for core/create in main is [2.0.1](https://github.com/Financial-Times/dotcom-tool-kit/blob/main/core/create/package.json#L3), I want to bump it up to 2.0.2 
- the prerelease works 

https://user-images.githubusercontent.com/20088576/167165188-3da1d38c-59c0-4672-ba45-6e7f617c689d.mov


apaleslimghost (changes requested):
**issue**: release-please looks at commits individually, so it needs to be the commit that actually changes a file in a package that has the `release-as:` i believe

apaleslimghost (approved):

PR #231, opened 2022-05-10
fix: read config file from cwd and use it when creating a new Mocha object 

JSRedondo:
Currently, we are not able to read the configuration file of Mocha tests.
This is specifically problematic when we need the --exit option to be set to true because otherwise, we have to use `process.exit()` on a global `after`, which is not that great regarding maintainability.

Other options may be necessary, for example, a specific timeout value, a coverage threshold, etc.


rowanmanning:
Looking over the Mocha repo I wonder whether we should be replicating all the ways to manage Mocha options? E.g. the built-in Mocha CLI also [supports loading from a `mocha` key in `package.json`](https://github.com/mochajs/mocha/blob/master/lib/cli/options.js#L176). Right now I can't see how it's possible to configure Mocha using Tool Kit but maybe we're missing something 😄  might be worth double-checking with Platforms whether we're missing something before writing too much more code though?

JSRedondo:
> Looking over the Mocha repo I wonder whether we should be replicating all the ways to manage Mocha options? E.g. the built-in Mocha CLI also [supports loading from a `mocha` key in `package.json`](https://github.com/mochajs/mocha/blob/master/lib/cli/options.js#L176). Right now I can't see how it's possible to configure Mocha using Tool Kit but maybe we're missing something 😄 might be worth double-checking with Platforms whether we're missing something before writing too much more code though?

Yep, for sure we need to check with Platforms, this is a delicate issue. On the other side, we have also the same problem with Eslint task, @alexmuller wrote me a comment in another PR about this.

PR #232, opened 2022-05-10
Cpp-882 Fix 'There are problems with your Tool Kit configuration'

serena97:
This error gets thrown when running `npm init @dotcom-tool-kit` and selecting both Eslint and Mocha, without being properly handled. This PR fixes this by dynamically importing hasToolKitConflicts to stop it from failing the instanceof check.

sidenote: before merging, will squash refactor into fix commit because I want to release the refactor commit as well after merging

### DIFF
--- a/core/create/package.json
+++ b/core/create/package.json
@@ -17,7 +17,7 @@
     "@financial-times/package-json": "^3.0.0",
     "@quarterto/parse-makefile-rules": "^1.1.0",
     "dotcom-tool-kit": "^2.1.1",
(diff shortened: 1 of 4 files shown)


apaleslimghost:
> squash refactor into fix commit because I want to release the refactor commit as well after merging

i don't think you'll need to do that, release-please releases everything on `main`

diegoFernandoMauroFT:
Looks good to me, as Kara commented, release-please should release everything on main on its own

PR #234, opened 2022-05-11
Cpp 617 add a pa11y plugin release fix

diegoFernandoMauroFT:


serena97 (commented):
I see there's 2 commits, we only need one of them

serena97 (approved):

PR #237, opened 2022-05-12
chore: release 2.0.2

serena97:
Release-As: 2.0.2


ivomurrell (commented):
This won't fix the problem, `create` has already been successfully released, the problem is release-please is trying to release it again. The release-please state needs to be fixed by manually creating release tags for missing components and deleting the `autorelease:pending` tag from the release-please PR.

PR #239, opened 2022-05-17
chore: fix types for pa11y

apaleslimghost:


serena97 (approved):

serena97 (changes requested):
file changes look good, but we should change this commit message to fix

apaleslimghost:
it's not really `fix` since the incorrect types weren't ever in any actually-released version, just the rebased `main`

serena97 (approved):

PR #245, opened 2022-06-01
CPP-851 improve plugin import messages

apaleslimghost:
to do @ivomurrell:

- [x] refactor the `Conflict<T>` and `Invalid<T>` types into (an equivalent of) `Either<L, R>`
- [x] have `validatePlugin` return instead of throwing for validation errors?
- [x] collect the validation errors with an equivalent of `sequence`
- [x] format and throw all the validation errors at once between `loadPlugin` and `resolvePlugin` in `loadConfig`
- [x] make the validation error messages specific as to what's gone wrong, e.g.:
  - these tasks etc from this plugin are not a compatible instance of `Task` etc
  - `tasks` from this plugin is not an array
  - an error was thrown when loading this plugin entry point


core/cli/src/plugin.ts
```diff
@@ -7,6 +7,7 @@ import {
   Invalid,
```
serena97, line 7: nitpick: Invalid isn't being used anymore and can be removed

serena97 (commented):
I've left some minor comments and it's looking good so far - I will run this tmrw locally and try to stimulate some errors to see whether there's any issues I can find

lib/types/src/index.ts
```diff
@@ -1,16 +1,78 @@
...
+
+    // objectToCheck from a plugin is compatible with this CLI if its version
+    // is semver-compatible with the @dotcom-tool-kit/types included
+    // by the CLI (which is what's calling this). so, prepend ^ to
+    // our version, and check our version satisfies that.
+
+    // this lets e.g. a CLI that includes types@2.2.0 load any plugin
+    // that depends on any higher minor version of types.
+    const range = `^${this.version}`
+    return semver.satisfies(objectToCheck.version, range)
```
serena97, line 52: suggestion: if this returns false, it would also be helpful to return a message like `expected ${objectToCheck.name} version to satisfy ${range}, but got ${objectToCheck.version}`so that users can see what version they need to install.

serena97:
question: should this be a breaking change?

core/cli/src/config.ts
```diff
@@ -23,17 +23,71 @@ export interface PluginOptions {
...
+  } else {
+    if (second.valid) {
+      return first
+    } else {
+      return { valid: false, reasons: [...first.reasons, ...second.reasons] }
+    }
+  }
+}
+
+export function sequenceValidated<T>(validated: Validated<T>[]): Validated<T[]> {
```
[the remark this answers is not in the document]
ivomurrell: It's referencing the [`sequence`](https://hackage.haskell.org/package/base-4.16.1.0/docs/Prelude.html#v:sequence) function from Haskell, though this is a specialisation of a much more generic function. `reduce` is a much better analogy in JavaScript so I'll update the name!

serena97 (approved):

PR #246, opened 2022-06-08
fix: eslint peerdep syntax

serena97:


ivomurrell:
😱
**todo**: would you be able to make a slightly more specific commit message so that the change is understandable in the `CHANGELOG` that release-please will generate? 😁

ivomurrell (approved):

PR #248, opened 2022-06-17
CPP-898 Allow mocha to read config files

ivomurrell:
We want to delegate as much configuration to tools themselves rather than exposing options via Tool Kit. Unfortunately, the mocha plugin was invoking mocha with a low-level API and skipping the code which reads configuration files, meaning that users could not set options like they can with plugins such as Jest and Webpack.

To solve this let's just invoke the mocha CLI directly rather than via a lower-level API so that it calls all the setup code it would for a normal invocation, including looking for configuration files. The user can also set the path for the config file explicitly in the Tool Kit options.

The files that mocha reads can now be set in its own configuration file so we don't really need to have them be set by Tool Kit. The option to do so might be removed in a future major release of the plugin.

### DIFF
--- a/lib/types/src/schema/mocha.ts
+++ b/lib/types/src/schema/mocha.ts
@@ -1,7 +1,8 @@
 import { SchemaOutput } from '../schema'
 
 export const MochaSchema = {
-  files: 'string'
+  files: 'string',
+  configPath: 'string?'
 } as const
 export type MochaOptions = SchemaOutput<typeof MochaSchema>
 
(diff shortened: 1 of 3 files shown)


serena97 (approved):
looks good to me! it's a shame that mocha api doesn't have a configpath option

ivomurrell:
There are ways to load config via the API (https://mochajs.org/api/module-lib_cli.html#.loadRc) but I there may be other weird behavioural quirks we find later on with using the low level API. I don't think invoking mocha programmatically is really the intended purpose of the API (based on the [introduction](https://mochajs.org/api/index.html))

PR #249, opened 2022-06-17
Allow arbitrary arguments to be set for node process in options

ivomurrell:
These could be either options for node like `--inspect` or arguments for the Node process you're starting. There's not a simpler way to do this seeing as there are no generalised configuration files applicable (unlike for `nodemon`.)

### DIFF
--- a/lib/types/src/schema/node.ts
+++ b/lib/types/src/schema/node.ts
@@ -1,7 +1,8 @@
 import { SchemaOutput } from '../schema'
 
 export const NodeSchema = {
-  entry: 'string?'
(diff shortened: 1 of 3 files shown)


serena97 (approved):
looks good to me! we'll need to document this so people will know how to put this in their toolkitrc.yml: 
```
  '@dotcom-tool-kit/node':
    args:
      - --inspect
```

ivomurrell:
Yeah, that will be part of [CPP-871](https://financialtimes.atlassian.net/browse/CPP-871)

PR #252, opened 2022-06-20
Write chosen port for nodemon too

ivomurrell:
The `node` plugin writes the port it chooses to the `.toolkitstate` which the `next-router` plugin can then read from and register with the `next-router` server. nodemon was missing the same behaviour so let's copy it over.


juanSanchezAlcala (approved):
Seems ok

diegoFernandoMauroFT (approved):
LGTM

PR #258, opened 2022-06-27
CPP-915 Allow Pa11y to use config defined in .pa11yci

ivomurrell:
Instead of passing static options to Pa11y, calling the `pa11y-ci` binary instead reads config defined in a `.pa11yci[.js[on]]` file. Seeing as this can be a `.js` file, the config can be dynamically generated, making the tooling much more flexible. This was how n-gage interfaced with Pa11y.

This also solves an issue that reported validation errors from Pa11y as internal errors within Pa11y.

### DIFF
--- a/lib/types/src/schema/pa11y.ts
+++ b/lib/types/src/schema/pa11y.ts
@@ -1,13 +1,7 @@
 import { SchemaOutput } from '../schema'
 
 export const Pa11ySchema = {
-  host: 'string?',
-  wait: 'number?',
-  tests: 'array.string?',
-  exceptions: 'array.string?',
-  hide: 'array.string?',
-  viewports: 'array.string?',
-  screenCapturePath: 'string?'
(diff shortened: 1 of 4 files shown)


joelcarr:
Looks good to me, just need to update this branch with `main` 🙌

joelcarr (approved):

joelcarr (changes requested):
**issue:** This looks like it should be a breaking change commit as we are changing how to set the config.

ivomurrell:
Good point: I initially thought it would be okay doing a minor update as those are treated as breaking changes in prerelease versions (`< 1.0.0`) but actually this change would have been released as a patch update because we have [`bump-minor-pre-major`](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md#configfile) enabled in our release-please config.

joelcarr (approved):
Brilliant!

PR #269, opened 2022-07-20
Include commits from #245 in release-please release

ivomurrell:
Due to https://github.com/googleapis/release-please/issues/1533, the changes in #245, aren't included in the generated release-please PR. I've made a few tweaks to the documentation in all the packages that were changed and appended the relevant commit headers in the footer of the messages, which release-please will interpret as new commit messages to include. This means that as well as calculating the version bumps correctly, the changes will also be listed in the CHANGELOG files and in the release notes.

### DIFF
--- a/core/create/CHANGELOG.md
+++ b/core/create/CHANGELOG.md
@@ -1,16 +1,16 @@
 # Changelog
 
-### Dependencies
+## [2.0.6](https://github.com/Financial-Times/dotcom-tool-kit/compare/create-v2.0.5...create-v2.0.6) (2022-07-12)
 
-* The following workspace dependencies were updated
-  * dependencies
-    * dotcom-tool-kit bumped from ^2.0.0 to ^2.1.0
 
 ### Dependencies
 
 * The following workspace dependencies were updated
   * dependencies
-    * dotcom-tool-kit bumped from ^2.2.0 to ^2.2.1
+    * dotcom-tool-kit bumped from ^2.2.2 to ^2.2.3
+
+## [2.0.5](https://github.com/Financial-Times/dotcom-tool-kit/compare/create-v2.0.4...create-v2.0.5) (2022-06-21)
+
 
 ### Dependencies
 
@@ -19,11 +19,14 @@
     * @dotcom-tool-kit/types bumped from ^2.2.0 to ^2.3.0
     * dotcom-tool-kit bumped from ^2.2.1 to ^2.2.2
 
+## [2.0.4](https://github.com/Financial-Times/dotcom-tool-kit/compare/create-v2.0.3...create-v2.0.4) (2022-06-08)
+
+
 ### Dependencies
 
 * The following workspace dependencies were updated
   * dependencies
-    * dotcom-tool-kit bumped from ^2.2.2 to ^2.2.3
+    * dotcom-tool-kit bumped from ^2.2.0 to ^2.2.1
 
 ## [2.0.3](https://github.com/Financial-Times/dotcom-tool-kit/compare/create-v2.0.2...create-v2.0.3) (2022-06-07)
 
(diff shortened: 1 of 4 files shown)


core/create/CHANGELOG.md
```diff
@@ -60,6 +63,15 @@
     * @dotcom-tool-kit/types bumped from ^2.0.0 to ^2.1.0
     * dotcom-tool-kit bumped from ^2.1.0 to ^2.1.1
```
serena97, line 64: question: do we have to add the part for 2.0.1 -> 2.1.0 as well?

core/create/CHANGELOG.md
```diff
@@ -60,6 +63,15 @@
...
 
+### [2.0.1](https://github.com/Financial-Times/dotcom-tool-kit/compare/create-v2.0.0...create-v2.0.1) (2022-05-05)
```
serena97, line 66: **praise**: nice spot - this must have come from here https://github.com/Financial-Times/dotcom-tool-kit/blob/7473c06517357a61932a28566e788a97ecec3992/core/create/CHANGELOG.md

**question**: should this be dated 2022-05-03 since the file added this part on 3rd of May?

core/create/CHANGELOG.md
```diff
@@ -1,16 +1,16 @@
...
-    * dotcom-tool-kit bumped from ^2.0.0 to ^2.1.0
 
 ### Dependencies
 
 * The following workspace dependencies were updated
   * dependencies
-    * dotcom-tool-kit bumped from ^2.2.0 to ^2.2.1
+    * dotcom-tool-kit bumped from ^2.2.2 to ^2.2.3
+
+## [2.0.5](https://github.com/Financial-Times/dotcom-tool-kit/compare/create-v2.0.4...create-v2.0.5) (2022-06-21)
```
serena97, line 12: question: should this be dated (2022-06-20) as well because that's the date of when release please added the dependencies section https://github.com/Financial-Times/dotcom-tool-kit/blob/4f3f9b422b0771b8cdf3b7664e3b4dbc38d3dbe9/core/create/CHANGELOG.md?

lib/types/src/index.ts
```diff
@@ -9,7 +9,7 @@ const packageJsonPath = path.resolve(__dirname, '../package.json')
 const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'))
 const version: string = packageJson.version
 
-// uses Symbol.for, not Symbol, so they're compatible across different
+// uses Symbol.for, not Symbol, so that they're compatible across different
```
serena97, line 12: question: with the additional commits you have added into the commit message footer - 
ie. 
```
feat: add validation errors to compatibility test in types package
refactor: better name than "something"
feat: loosen task and hook instance checks
```
will this mean release please will generate changelogs for these multiple features/refactor, from what I understand that are just a reference to the old commits that we've made in `main`? unfortunately the [doc](https://github.com/googleapis/release-please#what-if-my-pr-contains-multiple-fixes-or-features) doesn't make it very clear, nonetheless I don't think we want release-please to add these three to the changelog

serena97 (commented):
thanks for making the changes, you did a great job spotting the missing changelogs. I've got a few questions - once that's addressed I'm happy to approve

core/create/CHANGELOG.md
ivomurrell: when filling in the dates I chose to use the days we actually[ published the changes](https://github.com/Financial-Times/dotcom-tool-kit/releases/tag/create-v2.0.1) rather than the day when release-please generated the changelogs. it's not a big deal that it doesn't recalculate these (there's no way it could really) but if I'm going to add them manually I want to do it properly!

lib/types/src/index.ts
ivomurrell: like those docs say, release-please [parses](https://github.com/googleapis/release-please/blob/f75e49f6b751347456746d9c4068191d340c8e1e/src/commit.ts#L223-L288) all the text in the git footer that look like commit messages as if they are additional conventional commits. these three commits are old commits that have already been committed to main but were missed by release-please because of https://github.com/googleapis/release-please/issues/1533. these changes are all from #245 though so why wouldn't we want to include them in the changelog?

serena97: ah right, they're the commits that are missed, that makes sense - thanks for explaining!

serena97 (approved):

PR #273, opened 2022-08-04
fix: prevent infinite recursion when logging "["

apaleslimghost:
call the original, unwrapped `stream.write` instead of the wrapper, which was previously being called because of pass by reference and mutability


ivomurrell (approved):
this works great, thank you so much 😭

apaleslimghost:
fyi this isn't the same issue as CPP-937

ivomurrell:
whoops guess I didn't get around to creating a ticket for this bug, my bad

PR #276, opened 2022-08-08
CPP-942 Fix crash when trying to scale undeployed Heroku apps

ivomurrell:
# Description

We want to scale apps before promoting them to production as the preboot phase after promotion can take minutes and scaling is not possible during that time. However, scaling will fail if this is the first time the app has been deployed, so do the deployment first then. Hopefully (!), there shouldn't be a preboot phase if we don't need to switch from an old version either so it should be safe to scale the app immediately afterwards.

I've been able to test and confirm that it runs the original code path for next-static (with `appIds.length` totalling `1`) but it would be much harder to confirm the other path is working seeing as we don't have any Heroku pipelines without deployed apps currently. Instead, I've left a log in the code to message the Platforms team if that path is reached.

# Checklist:

### DIFF
--- a/plugins/heroku/src/tasks/production.ts
+++ b/plugins/heroku/src/tasks/production.ts
@@ -37,18 +37,42 @@ options:
         throw error
       }
 
-      for (const [appName, typeConfig] of Object.entries(scaling)) {
-        this.logger.verbose(`scaling app ${styles.app(appName)}...`)
-        for (const [processType, { quantity, size }] of Object.entries(typeConfig)) {
-          await scaleDyno(this.logger, appName, quantity, processType, size)
+      const scale = async () => {
+        for (const [appName, typeConfig] of Object.entries(scaling)) {
+          this.logger.verbose(`scaling app ${styles.app(appName)}...`)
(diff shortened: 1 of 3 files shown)


plugins/heroku/src/tasks/production.ts
```diff
@@ -37,18 +37,42 @@ options:
         throw error
       }
 
-      for (const [appName, typeConfig] of Object.entries(scaling)) {
-        this.logger.verbose(`scaling app ${styles.app(appName)}...`)
-        for (const [processType, { quantity, size }] of Object.entries(typeConfig)) {
-          await scaleDyno(this.logger, appName, quantity, processType, size)
+      const scale = async () => {
```
serena97, line 40: **praise**: +1 for separating them into functions

serena97 (approved):
I think this is a sensible workaround, and since this doesn't affect the happy path, and we don't have a beta release process, I think merging this is fine and it will make testing easier (as opposed to trying to adapt test-deployment.mts). We should definitely try to make time to test this soon by creating a new test app, I've started [one](https://github.com/Financial-Times/test-heroku/settings/installations) (and hopefully will have time to finish setting it up this week)

PR #283, opened 2022-08-17
Properly test if app has been deployed previously

ivomurrell:
# Description

We need different deployment behaviour for Heroku apps depending on whether they've been deployed before or not. Previously, we'd just check for whether an app existed in the pipeline, but this isn't a useful heuristic as apps are apparently created before a build is run. So let's instead call the Heroku API and see what the `slugSize` of an app is: this value appears to be `null` until the app has been build and promoted to production.

This does mean that we're making an extra API call before we deploy an app, but this should hopefully result in more reliable deployments.

# Checklist:

### DIFF
--- a/plugins/heroku/src/declarations.d.ts
+++ b/plugins/heroku/src/declarations.d.ts
@@ -28,6 +28,11 @@ declare module 'heroku-client' {
     stage: string
   }
 
+  export type HerokuApiResGetApp = {
+    id: string
+    slugSize: number | null
+  }
+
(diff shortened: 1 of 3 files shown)


joelcarr:
@ivomurrell is this still relevant?

ivomurrell:
> @ivomurrell is this still relevant?

Yes! I've rebased the changes onto `main` so they should be ready to review.

jkerr321 (approved):

PR #284, opened 2022-08-22
fix(upload-assets-to-s3): fix destination folder

juanSanchezAlcala:
# Description

The plugin upload-assets-to-s3 force to upload the assets with the same structure as "directory" property has ,this property indicates the folder to obtain the files. This situation is not convenient some times , for example in a project where the public folder is in a child project or in different folder that the root but you need to move this files directly to "destination" property without adding "directory" folders.

# Checklist:

### DIFF
--- a/plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
+++ b/plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
@@ -32,7 +32,7 @@ export default class UploadAssetsToS3 extends Task<typeof UploadAssetsToS3Schema
     const basename = file.split('/').splice(1).join('/') // remove first directory only
     const type = getFileType(basename)
     const encoding = getFileEncoding(basename)
-    const key = path.posix.join(options.destination, basename)
+    const key = options.destination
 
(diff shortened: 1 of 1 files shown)


ivomurrell:
Hey Juan, would you be able to give an example of the structures of the source directory in your project and how you'd want the destination directory to look? At the moment this change will mean that all files will be uploaded to the same S3 key, overwriting each other.

juanSanchezAlcala:
> Hey Juan, would you be able to give an example of the structures of the source directory in your project and how you'd want the destination directory to look? At the moment this change will mean that all files will be uploaded to the same S3 key, overwriting each other.

The example is [next-globetrotter](https://github.com/Financial-Times/next-globetrotter) , I haven't found a way to upload from parent project the content on packages/app/public to /hashed-assets/page-kit instead of /hashed-assets/page-kit/app/public. The assets in that project should be uploaded to /hashed-assets/page-kit to work properly when deploy to production or staging. About overwriting, i think we shouldn't be worried due the assets has the hash in the name and if I am not mistaken in the case that we are uploading a file with the same hash that would mean is exactly the same.

PR #285, opened 2022-08-23
CPP-954 Strip whole base directory from S3 upload key

ivomurrell:
# Description

Tool Kit has a plugin to upload assets within a folder to S3. You can set a directory option to specify where in your project the public assets are stored. The plugin is meant to strip out the directory part of the path from the destination key so that you aren’t forced to have a `public/assets/blah/` folder structure in S3 that mirrors your project’s directory structure. However, previously, only the first directory path is stripped from the key, so files would still have a path like `assets/blah/foo.txt`.

I think this is safe to release as a patch instead of a breaking change because the previous behaviour was surprising and I don't think anyone was relying on it.

Closes #284

# Checklist:


juanSanchezAlcala (approved):

package-lock.json
```diff
@@ -20581,7 +20581,7 @@
     },
     "plugins/pa11y": {
       "name": "@dotcom-tool-kit/pa11y",
-      "version": "0.3.2",
```
joelcarr, line 20584: **question:** Is this change required in this commit?

ivomurrell: this change happens automatically after running `npm install`. release-please isn't able to update the lockfiles when it merges releases so it will have to get updated manually instead. I'd consider it a minor change that we can smuggle into other PRs/commits to avoid us constantly manually updating it in dedicated PRs (and remembering to exclude it from staging in other unrelated commits.)

joelcarr (approved):

ivomurrell:
[I can't find](https://cs.github.com/?scopeName=All+repos&scope=&q=org%3AFinancial-Times+%40dotcom-tool-kit%2Fupload-assets-to-s3) any apps which were relying on the previous behaviour of only stripping the first folder in the path (every repository so far is just using the default `public` option which will be the folder that is stripped,) so should be safe to release this as a fix.

PR #287, opened 2022-08-25
fix: update PORT preference so the PORT set in Vault is not override

gyss:
# Description

At this moment, if you set a PORT env var in Vault /development (eg. [dotcom pages](https://vault.in.ft.com:8080/ui/vault/secrets/secret/show/teams/next/dotcom-pages/development)) the value doesn't have the correct preference and it uses the default port [3001, 3002, 3003].

In this PR we change the preference to make so we use:
- 1st: env var
- 2nd: Vault value
- 3rd: default values [3001, 3002, 3003]

# Checklist:

### DIFF
--- a/plugins/nodemon/src/tasks/nodemon.ts
+++ b/plugins/nodemon/src/tasks/nodemon.ts
@@ -39,8 +39,8 @@ export default class Nodemon extends Task<typeof NodemonSchema> {
     this.logger.verbose('starting the child nodemon process...')
 
     const env = {
-      ...vaultEnv,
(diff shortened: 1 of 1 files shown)


joelcarr:
Closing this PR for now as we [discussed](https://financialtimes.slack.com/archives/C02TRE2V2Q1/p1661502304238269?thread_ts=1661446013.943549&cid=C02TRE2V2Q1) as the current solution works. We will review this again if it turns out there is an issue that is caused by the port our Tool Kit script chooses.

PR #288, opened 2022-08-26
feat: add optional upload-assets-to-s3 plugin when migrating

alexmuller:
We've had an incident where an app didn't include the upload-assets-to-s3 plugin and its CSS and JS was unavailable in production. As this plugin is quite important (and used by most of our frontend apps) I think we should offer it during the init script.

# Checklist:

### DIFF
--- a/core/create/src/index.ts
+++ b/core/create/src/index.ts
@@ -118,7 +118,12 @@ async function mainPrompt() {
           },
           { title: 'ESLint', value: 'eslint', description: 'an open source JavaScript linting utility' },
           { title: 'Prettier', value: 'prettier', description: 'an opinionated code formatter' },
-          { title: 'lint-staged', value: 'lint-staged', description: 'run linters on git staged files' }
+          { title: 'lint-staged', value: 'lint-staged', description: 'run linters on git staged files' },
+          {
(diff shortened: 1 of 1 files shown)


andygout:
Does this PR include the upload-assets-to-s3 plugin in the `@dotcom-tool-kit/frontend-app` compilation plugin or is it doing something different?

And in the case of the latter is it worth doing the former?

alexmuller:
That's an interesting question, I don't know enough about Tool Kit to know about whether/why the frontend app plugin should or should not always depend on upload-assets-to-s3.

alexmuller:
Rowan points out [on Slack](https://financialtimes.slack.com/archives/C04030GAEGG/p1661524714818029?thread_ts=1661517549.902219&cid=C04030GAEGG):

> I don’t even think we could reasonably install it by default in the [Tool Kit front-end app plugin](https://github.com/Financial-Times/dotcom-tool-kit/blob/main/plugins/frontend-app/.toolkitrc.yml) because there are internal-facing front end apps which serve their assets locally rather than uploading them to hashed assets

ivomurrell (approved):
thanks for finding the perfect place to surface this plugin for people migrating!

PR #289, opened 2022-09-07
CPP-937 Avoid infinite loops when custom plugins don't have a config file

ivomurrell:
# Description

We've found that it's quite easy to forget to create a `.toolkitrc.yml` config file for your custom plugin (even though it's mentioned in the documentation). Omitting the config file will mean that cosmiconfig will try and traverse parent directories until it finds a valid config. This will likely be the config at the app's root instead, which will (at least transitively) include the current plugin in its `plugin` property, causing an infinite loop of config resolution.

Let's try and catch when we've resolved the config to the same root version and print a warning to remind people this can have unexpected results. This probably won't catch loops when one plugin installs another within the `node_modules` directory, but at that point it's likely going to be a plugin we own and have audited so should have a config file.

# Checklist:

### DIFF
--- a/core/cli/src/rc-file.ts
+++ b/core/cli/src/rc-file.ts
@@ -1,19 +1,41 @@
-import { cosmiconfig } from 'cosmiconfig'
+import { styles as s } from '@dotcom-tool-kit/logger'
 import { RCFile } from '@dotcom-tool-kit/types/src'
+import { cosmiconfig } from 'cosmiconfig'
+import * as path from 'path'
+import type { Logger } from 'winston'
 
 export const explorer = cosmiconfig('toolkit', { ignoreEmptySearchPlaces: false })
+const emptyConfig = { plugins: [], hooks: {}, options: {} }
+let rootConfig: string | undefined
 
 type RawRCFile = {
   [key in keyof RCFile]?: RCFile[key] | null
 }
 
-export async function loadToolKitRC(root: string): Promise<RCFile> {
-  const result = (await explorer.search(root)) as { config: RawRCFile | null } | null
-  if (!result || !result.config) return { plugins: [], hooks: {}, options: {} }
+export async function loadToolKitRC(logger: Logger, root: string, isAppRoot: boolean): Promise<RCFile> {
+  const result = await explorer.search(root)
 
+  if (!result?.config) {
+    return emptyConfig
+  }
+  if (isAppRoot) {
+    rootConfig = result.filepath
+  } else if (result.filepath === rootConfig) {
+    // Make sure that custom plugins which don't have a config file won't cause
+    // the resolver to use the root config instead and start an infinite loop
+    // of config resolution.
+    logger.warn(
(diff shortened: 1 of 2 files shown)


core/cli/src/rc-file.ts
```diff
@@ -1,19 +1,41 @@
...
+
+  if (!result?.config) {
+    return emptyConfig
+  }
+  if (isAppRoot) {
+    rootConfig = result.filepath
+  } else if (result.filepath === rootConfig) {
+    // Make sure that custom plugins which don't have a config file cause the
+    // resolver to use the root config instead and start an infinite loop of
+    // config resolution.
```
joelcarr, line 26: **issue:** I believe this description might be missing something. At the moment it reads to me that we use the root config and start an infinite loop. I think we mean that we use the root config instead of starting an infinite loop?

core/cli/src/rc-file.ts
```diff
@@ -1,19 +1,41 @@
...
-  if (!result || !result.config) return { plugins: [], hooks: {}, options: {} }
+export async function loadToolKitRC(logger: Logger, root: string, isAppRoot: boolean): Promise<RCFile> {
+  const result = await explorer.search(root)
+
+  if (!result?.config) {
+    return emptyConfig
+  }
+  if (isAppRoot) {
+    rootConfig = result.filepath
+  } else if (result.filepath === rootConfig) {
```
joelcarr, line 23: **question:** Could `rootConfig` here not be `undefined` to make this condition true?

core/cli/src/rc-file.ts
```diff
@@ -1,19 +1,41 @@
...
-  const result = (await explorer.search(root)) as { config: RawRCFile | null } | null
-  if (!result || !result.config) return { plugins: [], hooks: {}, options: {} }
+export async function loadToolKitRC(logger: Logger, root: string, isAppRoot: boolean): Promise<RCFile> {
+  const result = await explorer.search(root)
+
+  if (!result?.config) {
+    return emptyConfig
+  }
+  if (isAppRoot) {
+    rootConfig = result.filepath
```
joelcarr, line 22: **question:** What do we do with `rootConfig` here? It isn't referenced later on apart from on line 23 but as that is in an else condition it would have no context of this change… I think.

ivomurrell: `rootConfig` is stored at the module level so that it can be referenced across multiple calls to `loadToolKitRC`

core/cli/src/rc-file.ts
ivomurrell: `result.filepath` should never be `undefined` so that would `rootConfig` being `undefined` would never make this condition true

core/cli/src/rc-file.ts
joelcarr: Right of course! This function will be called multiple times in which we store the `rootConfig` in local memory for the subsequent calls.

joelcarr (approved):
🚫 ♾️ 🔁

PR #292, opened 2022-09-12
CPP-863 docs: add eslint readme.md

serena97:
# Description

Add eslint readme.md

# Checklist:

### DIFF
--- a/plugins/eslint/readme.md
+++ b/plugins/eslint/readme.md
@@ -1,0 +1,38 @@
+# @dotcom-tool-kit/eslint
+
+A plugin to run [eslint](https://eslint.org/) in your JavaScript code. This plugin uses the ESLint Node.js [API](https://eslint.org/docs/latest/developer-guide/nodejs-api) to run ESLint. It loads the ESLint configuration files (.eslintrc.* files) in your repo by default.
+
+## Installation
+
+With Tool Kit [already set up](https://github.com/financial-times/dotcom-tool-kit#installing-and-using-tool-kit), install this plugin as a dev dependency:
+
+```sh
+npm install --save-dev @dotcom-tool-kit/eslint
+```
+
+And add it to your repo's `.toolkitrc.yml`:
+
+```yaml
+plugins:
+    - '@dotcom-tool-kit/eslint'
+```
+
+## Options
+
+| Key | Description | Default value |
+|-|-|-|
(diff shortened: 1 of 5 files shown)


plugins/eslint/readme.md
```diff
@@ -0,0 +1,38 @@
...
+```yaml
+plugins:
+    - '@dotcom-tool-kit/eslint'
+```
+
+## Options
+
+| Key | Description | Default value |
+|-|-|-|
+| `files` | The glob patterns for lint target files. This can either be an array of string, or a string | `'**/*.js'` |
```
ivomurrell, line 24: **nitpick**: I've tried to make the description clearer
```suggestion
| `files` | The glob patterns for lint target files. This can either be a string or an array of strings | `'**/*.js'` |
```

plugins/eslint/readme.md
```diff
@@ -0,0 +1,38 @@
...
+| `config` | The [options](https://eslint.org/docs/latest/developer-guide/nodejs-api#-new-eslintoptions) for the ESLint constructor. This allows for additional flexibility as some of these options, such as `errorOnUnmatchedPattern`, are only applicable at this level but not in your eslintrc.* file | n/a |
```
ivomurrell, line 25: **thought**: maybe we should deprecate the `config` name as it has a different meaning to `config` in most other plugins (where it's the option to set the config path) and replace it with something like `options`?

plugins/eslint/readme.md
```diff
@@ -0,0 +1,38 @@
...
+Example:
+```
+'@dotcom-tool-kit/eslint':
+    files: server/*.js'
+    config:
+        errorOnUnmatchedPattern: false
+```
+## Tasks
+
+| Task | Description | Preconfigured hook |
```
ivomurrell, line 36: **thought**: should this section be called `Preconfigured hooks` if it can include multiple hooks...?

plugins/eslint/readme.md
serena97: I agree that we should rename it to `options` as the official [eslint docs](https://eslint.org/docs/latest/developer-guide/nodejs-api#-new-eslintoptions) calls it options anyways. I will include this change in a separate commit in this PR

ivomurrell (approved):
LGTM 🙌🏻

PR #295, opened 2022-09-21
fix: prettier plugin respects .prettierignore

apaleslimghost:
this (and the default value) are implemented in the Prettier CLI, not the API, so we should implement it in the task


apaleslimghost:
loooooooooooool https://github.com/Financial-Times/dotcom-tool-kit/blob/main/.prettierignore#L2

PR #298, opened 2022-10-07
Set region variable in heroku

jkerr321:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. 

# Checklist:


plugins/heroku/src/setConfigVars.ts
```diff
@@ -1,5 +1,6 @@
 import type { Logger } from 'winston'
 import heroku from './herokuClient'
+import type { HerokuApiResGetRegion } from 'heroku-client'
 import { ToolKitError } from '@dotcom-tool-kit/error'
 import { VaultEnvVars, Environment } from '@dotcom-tool-kit/vault'
 import { HerokuApiResPipeline } from 'heroku-client'
```
ivomurrell, line 6: **nitpick**: this import can be merged with the other import from `heroku-client`

```suggestion
import { ToolKitError } from '@dotcom-tool-kit/error'
import { VaultEnvVars, Environment } from '@dotcom-tool-kit/vault'
import type { HerokuApiResGetRegion, HerokuApiResPipeline } from 'heroku-client'
```

ivomurrell (approved):
This looks great! I'm glad we found a nice elegant fix to the problem in the end and these changes fit right in to the rest of the codebase 🎉  I left one trivial nitpick though I'd also suggest squashing all these commits into one `feat` commit before merging as most of the commits' changes have since been reverted in later commits.

plugins/heroku/src/setConfigVars.ts
jkerr321: Thanks Ivo I've done this on the branch and merged it into my one `feat` commit

PR #299, opened 2022-10-07
CPP-797 Share state between install hooks

ivomurrell:
# Description

Previously, every install method in a hook implementation would have to run independently of every other hook, even if other hooks would be writing to the same file. This was very complicated to manage and grossly inefficient, as, for instance, every CircleCI hook had to read and write to the same `.circleci/config.yml` file, one after the other. Some hooks also added dependencies on other hooks: if one CircleCI hook adds a filter to run on git tags, then the other hooks needed to follow suit so that they weren't ignored for a tag release. This meant a hardcoded `postInstall` function was required in the CLI to handle some post-processing depending on which hooks ran.

Now, state is able to be shared between hooks. A state object is passed between hooks who have the same `installGroup` property, e.g., all CircleCI hooks have the `'circleci'` `installGroup`. If state was set between hooks, then an additional `commitInstall` method is called after all the hooks in the group have run which will actually commit the generated state, e.g., write a `.circleci/config.yml` file based on the state.

I've tried to make these changes as backwards compatible as possible to reduce developer friction, and plugins using the old style of hooks should be compatible with these changes. However, updating to the new style of hooks will be a breaking change for a plugin, so the CircleCI and `package.json` hooks will be released with a new major version.

# Checklist:


joelcarr (commented):
**task:** We should add some documentation for how to use the new `installGroup` and `appendAfter` features for hooks.

lib/package-json-hook/src/package-json-helper.ts
```diff
@@ -1,31 +1,76 @@
...
+}
 
-export abstract class PackageJsonHelper extends Hook {
-	_packageJson?: PackageJson
-	abstract field: string
-	abstract key: string
-	abstract hook: string
+interface PackageJsonStateValue {
+  hooks: string[]
+  trailingString: string
```
joelcarr, line 15: **praise:** Much nicer name! It has clearer context of what the value is.

joelcarr (approved):
**praise:** Thanks for adding documentation for the new properties and helper function!

Looks great to me!

plugins/circleci/src/circleci-config.ts
```diff
@@ -32,64 +50,103 @@ export default abstract class CircleCiConfigHook extends Hook {
...
+      const formatMissingHooks = (hooks: string[]): string => hooks.map(styles.hook).join(', ')
       throw new ToolKitError(
-        `Please update your CircleCI config to include the \`${styles.hook(
-          this.job
-        )}\` job in the ${styles.heading('tool-kit')} workflow`
+        `Please update your CircleCI config to include the ${formatMissingHooks(
+          Object.keys(state.jobs)
+        )} jobs in the ${styles.heading('tool-kit')} workflow and the ${formatMissingHooks(
+          Object.keys(state.nightlyJobs)
+        )} jobs in the ${styles.heading('nightly')} workflow`
```
jkerr321, line 116: I have been running this locally with the tool-kit sandbox. When I remove the automated config message from the config.yml file and also remove one job from the workflow, e.g. `tool-kit/test`:
```yml
- tool-kit/test:
          node-version: 16.14-browsers
          requires:
            - tool-kit/build
          filters:
            tags:
              only: /^v\d+\.\d+\.\d+(-.+)?/
```

and then run `npx dotcom-tool-kit --install` I get the full error message indicating that all jobs are missing - is that the intended behaviour?

image

plugins/circleci/src/circleci-config.ts
```diff
@@ -1,29 +1,47 @@
...
+}
+
+const tagFilter = { filters: { tags: { only: `${semVerRegex}` } } }
+
+function isAutomatedConfig(config: YAML.Document): boolean {
+  const contents = config.contents as YAML.YAMLMap | undefined
+  const topKey = contents?.items[0].key as YAML.Scalar | undefined
+  const topComment = topKey?.commentBefore?.trim()
+  return topComment === automatedComment.slice(1).trim()
+}
```
jkerr321, line 27: This is much less readable than the previous check for automated config; `rawConfig.startsWith(automatedComment)`, why is this better?

I know you've added `YAML` to simplify the code elsewhere but you could still convert this check to be on a string 
 for simplicity: `config?.toString().startsWith(automatedComment)`

ivomurrell: Yeah, I think you're right that it would be more readable to just run a `toString()` conversion here. It's a reasonable compromise from this convoluted approach that was trying to improve upon the previous convoluted approach!

plugins/circleci/src/circleci-config.ts
```diff
@@ -102,15 +110,31 @@ export default abstract class CircleCiConfigHook extends Hook<CircleCIState> {
...
+            ? `the ${formatMissingHooks(missingJobs)} job(s) in the ${styles.heading('tool-kit')} workflow`
+            : ''
+        }${missingJobs.length > 0 && missingNightlyJobs.length > 0 ? ' and ' : ''}${
+          missingNightlyJobs.length > 0
+            ? `the ${formatMissingHooks(missingNightlyJobs)} job(s) in the ${styles.heading(
+                'nightly'
+              )} workflow`
+            : ''
+        }`
       )
```
joelcarr, line 138: **thought (non-blocking):** Can we move some of this logic into its' own method to reduce the size of this conditional block?

joelcarr (approved):
Looks good to me! The new refactor is definitely easier to understand.

core/cli/src/install.ts
```diff
@@ -1,6 +1,6 @@
 import { ToolKitError } from '@dotcom-tool-kit/error'
 import { OptionKey, setOptions } from '@dotcom-tool-kit/options'
-import groupBy from 'lodash.groupby'
+import groupBy from 'lodash/groupBy'
```
jkerr321, line 3: *praise:* This is cool! TIL about better tree shaking with lodash 🙌

jkerr321 (approved):
Thank you for addressing my comments 🥇

PR #303, opened 2022-11-02
docs: draft comments and suggestions for documentation improvements

jkerr321:
# Description

For discussion!

The problem:
- The existing README has a heavy focus on how flexible and configurable Tool Kit is but not on what an implementation is likely to look like across an app
- There are few concrete examples of how to use Tool Kit in the top level readme, rather the focus is on 'you _could_ use this, but you _could_ use that' which made me feel that I had to make decisions about every possible hook/task combo and felt overwhelming - in reality its usage is quite predictable and that should be made more clear
- It is unclear how many hooks and tasks there are or how they should (could) be used together
- Even once I'd used the migration tool I wasn't sure what was left to do - it wasn't clear to me which hooks and tasks hadn't yet been installed from the finite list of options, why some tasks had been applied to a given hook, and which hooks / tasks I might still want to install
- This is not helped by a lack of documentation on individual plugins
- What I now understand and would have liked communicated more clearly 
  - Most tasks are applied to a given hook by default - not every task can be applied to every hook and most tasks make sense to be applied to one or two hooks only
  - Most hooks are applied with the frontend or backend app plugin which is where most users will start with their installation. Once one of these plugins is installed migration is largely a case of installing tasks to work with those hooks rather than continuing to click through the code to understand what combination of hooks and tasks I still needed to apply. A list of possible tasks would have been helpful at this point.

The (potential) solution:
- Include some of the above in the main README in more or less words (I've started suggesting some changes to the README but actually they aren't as to the point as the stuff I've written above!)
- README's for each plugin including a list of exported tasks and hooks (examples for discussion included here)
- A document showing the full list of hooks and default tasks - [WIP document here](https://docs.google.com/spreadsheets/d/1YAWkQ7iV8Yq3S54O8RtjgSyrtCAnrNfRtyxCE26zpNI/edit?usp=sharing) - a separate page in the docs folder works best but I started with a sheet for ease
- A diagram (or something else) to explain the anatomy of an app development lifecycle, where hooks are applied at each point, and the tasks that can be used with them. I've chosen to use the frontend-app as an example - [WIP diagram here](https://lucid.app/lucidchart/b14b85ec-329a-4224-8a0b-0a60f2ed37f4/edit?viewport_loc=379%2C-1739%2C5331%2C3268%2C0_0&invitationId=inv_77fa8e7a-af28-4973-b2be-f7943c0f0f0e)

### DIFF
--- a/plugins/eslint/readme.md
+++ b/plugins/eslint/readme.md
@@ -17,6 +17,8 @@ plugins:
     - '@dotcom-tool-kit/eslint'
 ```
 
+running `npx dotcom-tool-kit --help` will show the hooks that Eslint has been applied to. By default these will be `test:local`, `test:ci`, `test:staged`.
+
 ## Options
 
 | Key | Description | Default value |
@@ -32,8 +34,14 @@ Example:
     options:
         errorOnUnmatchedPattern: false
 ```
-## Tasks
+## Exports Tasks
 
 | Task | Description | Preconfigured hooks |
 |-|-|-|
-| `Eslint` | runs `eslint` to lint and format target files | `test:local`, `test:ci`, `test:staged` |
+| `Eslint` | runs `eslint` to lint and format target files | `test:local`, `test:ci`, `test:staged` in the node plugin |
+
+### Exports Hooks
+none
+
+### How did n-gage do this?
+TBC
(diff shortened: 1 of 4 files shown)


joelcarr:
This fantastic review of our documentation has lead to new Jira Issues in the [CPP-996 epic](https://financialtimes.atlassian.net/browse/CPP-996) for us to work on.

jkerr321:
Closing this PR as actions have been taken from it and will be worked on in separate jira tickets and PRs

PR #306, opened 2022-11-09
chore: add tslib package to individual plugins

jkerr321:
Add tslib to individual plugins package.json files

So that packages don't need to rely on `tslib` being available from installation on another plugin. This avoids the error `Cannot find module 'tslib'` when running the migration tool.


ivomurrell (changes requested):
**issue**: `tslib` needs to be available at runtime so it should be in the `dependencies` field rather than `devDependencies`, the latter of which is pruned in production builds

ivomurrell (approved):
thanks for doing this fiddly work!

PR #307, opened 2022-11-09
Handle npm installation failing in create script

ivomurrell:
# Description

The `create` package has some special handling for when the error being thrown is relating to Tool Kit conflicts. To check this we need to pull in a package defining the types from the local project rather than from a local Tool Kit installation (or a different version from the registry), as otherwise some of the instanceof checks won't work. So we import our `error` package before printing any errors.

However, sometimes that package won't be available as something failed before we could even finish an npm installation. In those cases it's safe to assume that the issue isn't related to Tool Kit conflicts as we haven't even installed Tool Kit yet, so if we can't import the `error` package let's assert this isn't related to conflicts. This allows us to print the actual error, rather than obscuring it by an error about the `error` package being missing.

# Checklist:

### DIFF
--- a/core/create/src/index.ts
+++ b/core/create/src/index.ts
@@ -44,9 +44,19 @@ function installHooks(logger: typeof winstonLogger) {
   }).default(logger)
 }
 
-function hasToolKitConflicts(error: unknown) {
-  // we need to import hasToolkitConflicts from the app itself instead of npx or else hasToolkitConflicts and ToolKitConflictError will come from npx but error will come from the app level, leading to error failing the instanceof ToolKitConflictError check
-  return (importCwd('@dotcom-tool-kit/error') as typeof ToolkitErrorModule).hasToolKitConflicts(error)
+function hasToolKitConflicts(error: unknown): boolean {
+  try {
+    // we need to import hasToolkitConflicts from the app itself instead of npx
+    // or else hasToolkitConflicts and ToolKitConflictError will come from npx
+    // but the error will come from the app level, leading to the error failing
+    // the instanceof ToolKitConflictError check
+    const errorModule = importCwd('@dotcom-tool-kit/error') as typeof ToolkitErrorModule
+    return errorModule.hasToolKitConflicts(error)
+  } catch {
+    // if the error package isn't found then that probably means npm install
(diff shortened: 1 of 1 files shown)


core/create/src/index.ts
```diff
@@ -44,9 +44,19 @@ function installHooks(logger: typeof winstonLogger) {
...
+    // we need to import hasToolkitConflicts from the app itself instead of npx
+    // or else hasToolkitConflicts and ToolKitConflictError will come from npx
+    // but the error will come from the app level, leading to the error failing
+    // the instanceof ToolKitConflictError check
+    const errorModule = importCwd('@dotcom-tool-kit/error') as typeof ToolkitErrorModule
+    return errorModule.hasToolKitConflicts(error)
+  } catch {
+    // if the error package isn't found then that probably means npm install
+    // has failed and we should be logging out that error
+    return false
```
joelcarr, line 58: **question:** This function is just to determine how to log an error in a catch block therefore returning `false` here will still log the error. Have I understood this correctly?

jkerr321, line 58: Returning false here will [invoke this line](https://github.com/Financial-Times/dotcom-tool-kit/blob/f67f46fb83074ffa602cbfa6e2572cd424a7f90c/core/create/src/index.ts#L651) and therefore the winston will log the error and the process will exit?

jkerr321: [This block[(https://github.com/Financial-Times/dotcom-tool-kit/blob/f67f46fb83074ffa602cbfa6e2572cd424a7f90c/core/create/src/index.ts#L86-L92) is where that function is called so returning false will run the `else` block which will log the actual error rather than the `finished installing hooks, but found conflicts` error:

```js
if (hasToolKitConflicts(error)) {
      logger.log(id, { status: 'done', message: 'finished installing hooks, but found conflicts' })
    } else {
      logger.log(id, {
        status: 'fail',
        message: labels.fail,
        error: loggerError.logged ? undefined : loggerError
      })
    }
```

jkerr321 (approved):
Cool! This has also been a good opportunity for me to review this mammoth file, thanks!

joelcarr (approved):

PR #313, opened 2022-11-14
fix(sandbox): resolve hook conflict

jkerr321:
# Description
On running `npx dotcom-tool-kit --install` in the sandbox the install fails because there are hook conflicts on run:local. This PR configures run:local in the sandbox .toolkit.rc file to get past this error.

image

# Checklist:

### DIFF
--- a/core/sandbox/.toolkitrc.yml
+++ b/core/sandbox/.toolkitrc.yml
@@ -10,6 +10,9 @@ plugins:
   - '@dotcom-tool-kit/pa11y'
 
 hooks:
+  run:local:
+    - Nodemon
+    - NextRouter
   git:precommit:
     - SecretSquirrel
     - LintStaged


joelcarr (commented):
**thought:** I may be wrong - if a `.toolkitrc.yml` has modified a hook, I believe we have to define everything for that hook as it will not be merged to the other defaults. Does this still run the other local tasks like `NextRouter`?

jkerr321:
Yes Joel, I think you're right. I didn't think about that. I'll add `NextRouter`, I don't think I need I need the node tasks with WebpackWatch, is that right?
```
  run:local:
    - NextRouter
    - WebpackWatch
```

ivomurrell (approved):
I make the same change to sandbox whenever I'm testing locally!

PR #316, opened 2022-12-05
CPP-1034: fix(UploadAssetsToS3): handle AWS keys correctly

jkerr321:
# Description

The UploadAssetsToS3 class has an `accessKeyId` option. This may be provided via the `.toolkitrc` file and will be interpolated as a string. Therefore we need additional handling in the task to link the string variable name passed into the corresponding `process.env` variable value.

# Checklist:

- [x] My commit messages are [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), fr example: `feat(circleci): add support for nightly workflows`, `fix: set Heroku app name for staging apps too`


plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
```diff
@@ -15,8 +15,8 @@ export default class UploadAssetsToS3 extends Task<typeof UploadAssetsToS3Schema
   static description = ''
 
   static defaultOptions: UploadAssetsToS3Options = {
-    accessKeyId: process.env.aws_access_hashed_assets || '',
-    secretAccessKey: process.env.aws_secret_hashed_assets || '',
+    accessKeyId: 'aws_access_hashed_assets' || '',
+    secretAccessKey: 'aws_secret_hashed_assets' || '',
```
joelcarr, line 19: **nitpick:** We can remove the `|| ''` from these lines as they were there to set a default empty string is `process.env.[key]` did not exist.

plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
```diff
@@ -84,8 +84,8 @@ export default class UploadAssetsToS3 extends Task<typeof UploadAssetsToS3Schema
     const files = glob.sync(globFile, { cwd: options.directory })
 
     const s3 = new aws.S3({
-      accessKeyId: options.accessKeyId,
-      secretAccessKey: options.secretAccessKey
+      accessKeyId: process.env[options.accessKeyId],
+      secretAccessKey: process.env[options.secretAccessKey]
```
ivomurrell, line 88: **issue**: these option names are confusing now that they're meant to represent the name of the environment variables that store the access key, rather than the access key itself.

**suggestion**: I think we should deprecate these options (as opposed to deleting them) and replace them with options like `accessKeyIdEnvVar` and `secretAccessKeyEnvVar`. the option values can fall back to the values set in `accessKeyId` and `secretAccessKey` if they aren't set themselves to maintain backwards compatibility

ivomurrell (approved):
looks GREAT to me

PR #317, opened 2022-12-05
CPP-751 Still run provision step for nori/renovate builds

ivomurrell:
# Description

I think this was always an oversight rather than something we can only do now that we've made other improvements to the code.

The job now requires the `waiting-for-approval` job to have been run first, though this only runs for nori/renovate branches. If that job is filtered out then this job will instead simply run after `tool-kit/setup` is completed. n-gage would [previously](https://github.com/Financial-Times/next-user-facing-app-template/blob/bf8828c6fbdef6f7da706378c9dab9d5bef781c7/.circleci/config.yml#L189) create a different workflow for nori/renovate, with the `waiting-for-approval` job required before anything else was run. Tool Kit's approach is much simpler and reduces duplication.

# Checklist:

### DIFF
--- a/plugins/circleci-heroku/src/index.ts
+++ b/plugins/circleci-heroku/src/index.ts
@@ -4,8 +4,8 @@ import { TestCI } from '@dotcom-tool-kit/circleci/lib/index'
 export class DeployReview extends CircleCiConfigHook {
   job = 'tool-kit/heroku-provision'
   jobOptions = {
-    requires: ['tool-kit/setup'],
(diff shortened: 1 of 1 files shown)


joelcarr (approved):
Tested this change on `next-newsletters-page` by [making the same change to the CircleCI config.](https://github.com/Financial-Times/next-newsletters-page/commit/23b92bd96919bc1014a61bc6fa5d85dfc3bf5382) Here are some screenshots of it working in a branch prefixed with `renovate-` and a branch not prefixed with it:

image
image

PR #318, opened 2022-12-07
chore: add editor config to gitignore

jkerr321:
# Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ node_modules
 *.tsbuildinfo
 /.toolkitstate
 coverage
+/.vscode


ivomurrell:
imo user-specific files (I don't use VS Code for instance) that have been generated should be put into your personal `.git/info/exclude` ignore list rather than the shared `.gitignore` file https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#excluding-local-files-without-creating-a-gitignore-file

jkerr321:
> imo user-specific files (I don't use VS Code for instance) that have been generated should be put into your personal `.git/info/exclude` ignore list rather than the shared `.gitignore` file https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#excluding-local-files-without-creating-a-gitignore-file

Ahh Ivo I knew you would have a solution for this! TIL .git/info/exclude ignore list. Thank you!

PR #319, opened 2022-12-07
CPP-1020: Clearer circle install messaging

jkerr321:
# Description

Depends on https://github.com/Financial-Times/dotcom-tool-kit/pull/299.

Currently, if a user is missing specific Tool Kit jobs in their config, a [tailored error is thrown here](https://github.com/Financial-Times/dotcom-tool-kit/commit/5289193b88ed6dab6a459127bba1ee87ee27286f#diff-f157d7392cc3db08e0c5b213317a296767e475221a5be30ae3bcc8b94fa7613fR120) with the hooks they need to install. However, if a user is missing those hooks because their config is legacy config that hasn't yet been moved to Tool Kit (e.g. because they are migrating to Tool Kit for the first time), it is better to throw an error that explains how to auto generate Tool Kit config in the first place.

### Output:
image

# Checklist:


plugins/circleci/src/circleci-config.ts
```diff
@@ -115,6 +117,19 @@ export default abstract class CircleCiConfigHook extends Hook<CircleCIState> {
...
+
+    if (rawConfig && isNotToolKitConfig(rawConfig)) {
+      throw new ToolKitError(
+        `Your project has an existing CircleCI config file which doesn't contain a ${styles.heading(
+          'tool-kit'
+        )} workflow.
+If you would like a Tool Kit configured CircleCI config file to be generated for you please delete your existing CircleCI ${styles.filepath(
+          'config.yml'
+        )} and re-run:
+${styles.code('npx dotcom-tool-kit --install')}`
```
ivomurrell, line 129: **nitpick**: imo the newline here makes the suggestion harder to follow, especially as code snippets are already greyed out. seeing as it's just a single terminal invocation I think it's okay to be written inline

```suggestion
        )} and re-run ${styles.code('npx dotcom-tool-kit --install')}`
```

ivomurrell (approved):
this move towards making Tool Kit more understandable is v appreciated!

PR #320, opened 2022-12-07
CPP-1037 Reduce noise created by metadata prefixes

ivomurrell:
# Description

The Tool Kit logger prefixes each log message with labels to give context about where the message is coming from, e.g., which Tool Kit task or subprocess it's from. Some subprocesses (e.g., Jest) are prone to frequent stream flushing, causing their logs to be repeatedly interspersed with these labels greatly hindering readability. Let's work around this with a global variable that stores the last label we printed and only print another prefix if the label has changed, i.e., when the context for the message has changed.

# Checklist:

### DIFF
--- a/lib/logger/src/format.ts
+++ b/lib/logger/src/format.ts
@@ -1,6 +1,12 @@
 import winston from 'winston'
 import { styles } from './styles'
 
+// This global variable is used to prevent the same label being printed
+// multiple times in a row. Some processes (e.g., Jest) are prone to frequent
+// stream flushing, causing their logs to be repeatedly interspersed with the
+// labels, greatly hindering readability.
+let lastLabel: string | undefined
+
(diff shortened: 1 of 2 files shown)


joelcarr (approved):
🎉 Going to look great! Would be great to add a before an after screenshot to this PR for reference in the future.

jkerr321:
Looks good, thanks Ivo! I'll let JC approve once the screen shot is there 👍

ivomurrell:
original
output of npm test command before the PR changes have been applied, showing how labels for the JestLocal task are splitting out the tables Jest prints out for code coverage
new
output of npm test command after the PR changes have been applied, where the JestLocal labels are now less overbearing

Note how each row of the table doesn't print a new prefix in the new version, but changing the context will – in this case changing from info to warning logs and back (this is from Jest logging to both stdout and stderr)

PR #323, opened 2022-12-09
CPP-1027: add provision job to nightly workflow

joelcarr:
# Description

Adding the `tool-kit/heroku-provision` job to the Nightly workflow. Our heroku apps would provision a review app as part of the Nightly scheduled workflow when the CircleCI config was not automated. We are adding the job back for our automated by Tool Kit configs.

This also includes a removing being able to set `filters` job options to Nightly workflow jobs. This is because we configure in CircleCI that our nightly workflows are triggered against the `main` branch only. Therefore if we add a job to be in the Nightly workflow, it shouldn't need to filter which branches it won't/will run against. And in the case for the `tool-kit/heroku-provision` job where it has a filter not to run against the `main` branch for our `tool-kit` workflow, we need to cancel this option for the `nightly` workflow.

# Checklist:

### DIFF
--- a/plugins/circleci/src/circleci-config.ts
+++ b/plugins/circleci/src/circleci-config.ts
@@ -13,7 +13,7 @@ const majorOrbVersion = '2'
 
 interface CircleCIState {
   jobs: Record<string, JobConfig>
-  nightlyJobs: Record<string, JobConfig>
+  nightlyJobs: Record<string, Omit<JobConfig, 'filters'>>
   runOnVersionTags: boolean
   additionalFields: Record<string, unknown>
 }
@@ -106,7 +106,11 @@ export default abstract class CircleCiConfigHook extends Hook<CircleCIState> {
 
     state.jobs[this.job] = this.jobOptions
     if (this.addToNightly) {
-      state.nightlyJobs[this.job] = this.jobOptions
+      const jobOptionsWithoutFilters = {
+        ...this.jobOptions,
+        filters: undefined
+      }
+      state.nightlyJobs[this.job] = jobOptionsWithoutFilters
     }
     if (this.runOnVersionTags) {
       state.runOnVersionTags = true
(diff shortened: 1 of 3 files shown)


joelcarr:
A question for @Financial-Times/platforms: for commit https://github.com/Financial-Times/dotcom-tool-kit/commit/ce16686, should this be a breaking change? Technically it is removing a feature however it will not break the experience for the consumer.

plugins/circleci/src/circleci-config.ts
```diff
@@ -106,7 +106,11 @@ export default abstract class CircleCiConfigHook extends Hook<CircleCIState> {
 
     state.jobs[this.job] = this.jobOptions
     if (this.addToNightly) {
-      state.nightlyJobs[this.job] = this.jobOptions
+      const jobOptionsWithoutFilters = {
+        ...this.jobOptions,
+        filters: undefined
+      }
+      state.nightlyJobs[this.job] = { ...jobOptionsWithoutFilters }
```
ivomurrell, line 113: **question**: why do we clone `jobOptionsWithoutFilters` (by spreading it) again here?

plugins/circleci/src/circleci-config.ts
```diff
@@ -13,7 +13,7 @@ const majorOrbVersion = '2'
 
 interface CircleCIState {
   jobs: Record<string, JobConfig>
-  nightlyJobs: Record<string, JobConfig>
+  nightlyJobs: Record<string, Omit<JobConfig, 'filters'>>
```
ivomurrell, line 16: **praise**: lovely type transformation!

ivomurrell (commented):
Another great spot that came from dogfooding!

plugins/circleci/src/circleci-config.ts
joelcarr: Good question! I got in the flow of spreading. I will remove it now 👍

ivomurrell (approved):

PR #324, opened 2022-12-09
CPP-1018: add secret squirrel to bootstrap plugins

joelcarr:
# Description

Add `@dotcom-tool-kit/secret-squirrel` as a dependency to our bootstrap plugins, `@dotcom-tool-kit/component` and `@dotcom-tool-kit/backend-app`. Adding the plugin as part of the migration (then into the template apps) should bring Tool Kit more align to feature parity to `n-gage`. We want `secret-squirrel` to be installed by default but, like n-gage, it is possible to remove it from an application.

# Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20059,7 +20059,9 @@
       "license": "ISC",
       "dependencies": {
         "@dotcom-tool-kit/circleci-npm": "^3.0.0",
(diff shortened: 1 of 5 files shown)


jkerr321 (approved):
🚀

PR #326, opened 2022-12-12
CPP-1025 Add migration option to switch to scheduled CircleCI pipelines

ivomurrell:
# Description

CircleCI has [deprecated](https://circleci.com/docs/migrate-scheduled-workflows-to-scheduled-pipelines/) scheduled workflows and have replaced them with scheduled pipelines, and Tool Kit consequently only supports the latter. We found that many teams were missing this change during migration and as a result no longer have their nightly workflow running every night. To avoid this in the future let's highlight that the migration is necessary in the migration script, as well as offer to set up the new workflow programmatically via the CircleCI API.

Here's how the new prompt looks:
terminal prompt offering to migrate to CircleCI's scheduled pipelines, with all questions filled out and reporting success

I also refactored the package into multiple files so that it's a bit more parsable and modular hopefully, as well as allowing us to add more prompts in the future without expanding the file length even more.

Uh **_update_** whilst writing this PR: it looks like the deprecation has been [postponed](https://discuss.circleci.com/t/scheduled-pipelines-are-here/41684) as of last week. I'm not sure whether this should affect us though as this doesn't particularly justify supporting scheduled workflows in Tool Kit.

# Checklist:


core/create/src/prompts/main.ts
```diff
@@ -64,14 +63,11 @@ export default async ({
       {
         name: 'deleteConfig',
         // Skip prompt if CircleCI config doesn't exist
-        type: await fs
-          .access(circleConfigPath)
-          .then(() => 'confirm' as const)
-          .catch(() => null),
+        type: originalCircleConfig ? ('confirm' as const) : null,
         // This .relative() call feels redundant at the moment. Maybe we can just
         // hard-code the config path?
```
jkerr321, line 68: **praise:** nice!
**chore:** you can remove this comment now too :)

ivomurrell: omg I spent minutes yesterday trying to decide whether I should simplify the code by hardcoding that path without ever seeing that comment of past me also thinking it would make sense 😄

core/create/src/prompts/confirmation.ts
```diff
@@ -0,0 +1,42 @@
...
+  configFile: string
+}
+
+export default ({
+  deleteConfig,
+  addEslintConfig,
+  packagesToInstall,
+  packagesToRemove,
+  configFile
+}: ConfirmationParams): Promise<prompt.Answers<'confirm'>> =>
```
jkerr321, line 18: **suggestion:** I think it would be nice to wrap this function in curly braces. It is one return statement but it takes up many lines and I automatically looked for the encapsulating curly braces to see where the function ends. I know it's fairly unlikely but I also think it would be useful to wrap in curly braces in case anything another function is added to this file to more clearly separate them 

```ts
}: ConfirmationParams): Promise<prompt.Answers<'confirm'>> => {
    return prompt({
    name: 'confirm',
    type: 'confirm',
    message: () => {
      return `so, we're gonna:
    ....
    ....
    })
}

```

jkerr321 (approved):
This is so great! Splitting out the code makes the index file so much more readable, and the scheduled pipeline work is lovely!

joelcarr (approved):
🥳

PR #333, opened 2022-12-14
fix(upload-assets-to-s3): only upload files

jkerr321:
the upload-assets-to-s3 plugin works by looping over the files passed to it and uploading each to s3. When the files to upload includes a folder (with files inside) the directory itself is passed to s3 for upload.
s3 throws an error because it can only read files and not directories. This change updates the glob pattern of files to upload to not include directories.
The files within the directory are still uploaded successfully

# Description

**Before:**
list of files to upload includes `v2` which is a directory:
image

error is thrown (although files are still uploaded successfully):
image
image

**After:**
list of files to upload doesn't include any directories: 
image

files within the directory are still uploaded successfully:
image

top level:
image

inside `v2` directory:
image

# Checklist:


ivomurrell (approved):
perfect, thank you!

PR #341, opened 2022-12-21
CPP-871: docs(node): add readme

jkerr321:

### DIFF
--- a/plugins/node/readme.md
+++ b/plugins/node/readme.md
@@ -1,0 +1,34 @@
+# @dotcom-tool-kit/node
+
+Run your Node application.
+
+This plugin will be installed as a dependency of the [frontend-app](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/frontend-app), [backend-app](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/backend-app) plugins so you do not need to install it separately if you are using either of those plugins.
+
+## Installation
+
+Install `@dotcom-tool-kit/node` as a `devDependency` in your app:
+
+```sh
+npm install --save-dev @dotcom-tool-kit/node
+```
+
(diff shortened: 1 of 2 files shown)


plugins/node/readme.md
```diff
@@ -0,0 +1,34 @@
+# @dotcom-tool-kit/node
+
+Run your Node application.
```
jkerr321, line 3: **question:** is this verbose enough? It probably is... but it could be something like "Spawn a node process to run your application" but I thought the extra words might add confusion rather than clarity!

plugins/node/readme.md
```diff
@@ -0,0 +1,34 @@
...
+
+This plugin will be installed as a dependency of the [frontend-app](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/frontend-app), [backend-app](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/backend-app)  plugins so you do not need to install it separately if you are using either of those plugins.
```
joelcarr, line 5: **nitpick (non-blocking):** There is an extra space.

```suggestion
This plugin will be installed as a dependency of the [frontend-app](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/frontend-app), [backend-app](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/backend-app) plugins so you do not need to install it separately if you are using either of those plugins.
```

plugins/node/readme.md
joelcarr: I can't think of anything else to add so…

**praise:** Great short clear explanation without fluff!

plugins/node/readme.md
joelcarr, line 5: **praise:** This is a great addition to let teams know they might already have it installed for their app 🙌

joelcarr (approved):

PR #343, opened 2022-12-21
CPP-1042: fix lodash dependency

joelcarr:
In 454f9cd the import statements where updated to use the singular lodash module however it seems the cli package was missed when updating the dependencies. This means that when running `npx dotcom-tool-kit`, the `lodash/groupby` module was not found as our package.json list the dependency as `lodash.groupby`.

# Checklist:


ivomurrell (approved):
must have lost this in a merge conflict or something!? thanks for catching it!

PR #329, opened 2022-12-13
CPP-910 Bump release-please to v15

ivomurrell:
# Description

We are jumping from v13 to v15. [v14](https://github.com/googleapis/release-please/releases/tag/v14.0.0) was breaking because it dropped support for Node 12, which we don't use, and [v15](https://github.com/googleapis/release-please/releases/tag/v15.0.0), which just released yesterday, and doesn't seem to make any breaking changes that would affect us.

We previously had an [issue](https://financialtimes.atlassian.net/browse/CPP-908) where newer versions of release-please would be incorrectly trying to create releases for packages that didn't need updating. I tested to see whether this applied to v15 too and couldn't reproduce the issue, though we currently already have 29 updates due.

# Checklist:

- [x] My branch has been rebased onto the latest commit on main (don't merge main into your branch)
- [x] My commit messages are [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), for example: `feat(circleci): add support for nightly workflows`, `fix: set Heroku app name for staging apps too`

PR #345, opened 2023-01-04
Fix issues with release-please configuration

ivomurrell:
# Description

These changes should improve the behaviour of release-please so that

1. new components are not incorrectly assigned a tag of `v2.0.0` when initially released
2. changelogs only include changes relevant to each component, an issue we were seeing since we [updated](https://github.com/Financial-Times/dotcom-tool-kit/pull/329) to release-please `v15.0.0`.

Further details can be found in the commit descriptions. I have tested the behaviour by running release-please locally.

# Checklist:

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -1,5 +1,5 @@
 {
-  "bootstrap-sha": "bc6f822e7a6927b65b63c91021ede10dad73006f",
+  "bootstrap-sha": "47035e3bd4e3ac399319b3421ce86a82f1c1ab21",
   "plugins": ["node-workspace"],
   "changelog-type": "github",
   "prerelease": true,
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -1,7 +1,6 @@
 {
   "bootstrap-sha": "47035e3bd4e3ac399319b3421ce86a82f1c1ab21",
   "plugins": ["node-workspace"],
-  "changelog-type": "github",
   "prerelease": true,
   "bump-minor-pre-major": true,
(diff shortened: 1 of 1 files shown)


release-please-config.json
```diff
@@ -1,7 +1,6 @@
 {
-  "bootstrap-sha": "bc6f822e7a6927b65b63c91021ede10dad73006f",
+  "bootstrap-sha": "47035e3bd4e3ac399319b3421ce86a82f1c1ab21",
```
joelcarr, line 2: **question(non-blocking):** Shall we remove this line after the next release [as mentioned in the docs](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md#starting-commit)?

joelcarr (approved):
Lovely! One small question of no but non blocking.

release-please-config.json
ivomurrell: No, I think it's still used if it can't find a release for a new component, e.g., it will use that as the search limit when we add `plugins/typescript` for the first time. I assume if it's removed entirely then it will instead search the last 500 commits which seems to be its default limit. That's bad for us because the commit which sets every component to `v2.0.0` is included in that batch of the last 500 commits (for now)!

ivomurrell: So if we removed it after the next release I _think_ we'd run into the same issue the next time we create a new plugin.

joelcarr: Makes sense! Now we have this explainer bound to the commit's PR so it isn't removed by someone like me in the future 😄

PR #352, opened 2023-01-13
CPP-1022 Update Tool Kit orb to use change-api's 1.0 orb release

ivomurrell:
# Description

The Edge Delivery & Observability team updated the change-api orb to be a bit more streamlined and [released a new major version](https://github.com/Financial-Times/change-api-orb/releases/tag/v1.0.0). The main interesting change for us the removal of the dependency on `jq`, which means we don't have to run a command in our `tool-kit/setup` job to install `jq`, which was redundant if you weren't calling the Change API in a given workflow. Updating to the latest version also means we can get future feature releases automatically.

# Checklist:


jkerr321 (approved):
Nice! 👯

PR #354, opened 2023-01-19
Use Zod for declaring option schemas

ivomurrell:
# Description

I've been working on moving Tool Kit to use Zod over the last few 10% days. This is the result! [zod](https://zod.dev) is a schema library that I am now using to parse and validate the options that users set in Tool Kit config files. The features that Zod provides us are

1. validation of options that are passed – we're currently assuming that any options are of the correct type and that they exist if they are required which can lead to confusing errors if these assumptions are false. we'd sometimes try and work around this by checking that options were defined in a Task's `run()` method, even though the type wasn't given as `undefined`. this was easy to forget and created a lot of boilerplate code to give helpful error messages that can now be generated automatically from Zod errors via [`zod-validation-errors`](https://github.com/causaly/zod-validation-error).
2. standardising the way we specify object schemas – we previously supported the specification of options using a custom DSL declared via string literals. these strings would then be converted to TypeScript types using a special conditional mapping type which was fun to write but was inscrutable and inextensible. zod's schema definitions, on the other hand uses classes constructed via simple function calls that are well specified in their API and mirror TypeScript's type system. it also allows us to define more complex option types, like the [nested object used to specify Heroku pipelines](https://github.com/Financial-Times/dotcom-tool-kit/pull/354/files#diff-a6a41016c972b91e7774358c85a44aa966e6dff999deca90ea8b5638214e82d9R4-R11).
3. handling of default options – Zod [allows default values to be specified](https://zod.dev/?id=default) when options are undefined in the configuration. this allows us to remove the `defaultOptions` static field from tasks which was previously separated from the schema specification itself. this separation meant that even though a option might be optional in the schema, it would always be defined in reality if it had a value specified in `defaultOptions` and thus null assertions (`!`) would be required. we can now also get the default values for schemas and print them when users are being guided through options configuration in the `@dotcom-tool-kit/create` script.

These changes shouldn't be breaking, as we already assumed that the options in a configuration file were well-formed.

# Checklist:

### DIFF
--- a/core/create/src/prompts/options.ts
+++ b/core/create/src/prompts/options.ts
@@ -1,17 +1,18 @@
 import { rootLogger as winstonLogger, styles } from '@dotcom-tool-kit/logger'
 import type { RCFile } from '@dotcom-tool-kit/types'
-import type { Schema, SchemaPromptGenerator, SchemaType } from '@dotcom-tool-kit/types/src/schema'
+import type { PromptGenerators } from '@dotcom-tool-kit/types/src/schema'
 import type { ValidConfig } from 'dotcom-tool-kit/lib/config'
 import { promises as fs } from 'fs'
 import * as yaml from 'js-yaml'
 import partition from 'lodash/partition'
 import prompt from 'prompts'
+import { z } from 'zod'
 import type { Logger } from '../logger'
 
 async function optionsPromptForPlugin(
   toolKitConfig: RCFile,
   plugin: string,
-  options: [string, SchemaType][]
+  options: [string, z.ZodTypeAny][]
 ): Promise<boolean> {
   let pluginCancelled = false
   const onCancel = () => {
@@ -19,8 +20,8 @@ async function optionsPromptForPlugin(
   }
 
   for (const [optionName, optionType] of options) {
-    switch (optionType) {
-      case 'string':
+    switch (optionType._def.typeName as z.ZodFirstPartyTypeKind) {
+      case 'ZodString':
         const { stringOption } = await prompt(
           {
             name: 'stringOption',
@@ -33,7 +34,7 @@ async function optionsPromptForPlugin(
           toolKitConfig.options[plugin][optionName] = stringOption
         }
         break
-      case 'boolean':
+      case 'ZodBoolean':
         const { boolOption } = await prompt(
           {
             name: 'boolOption',
@@ -46,7 +47,7 @@ async function optionsPromptForPlugin(
           toolKitConfig.options[plugin][optionName] = boolOption
         }
         break
-      case 'number':
+      case 'ZodNumber':
         const { numberOption } = await prompt(
           {
             name: 'numberOption',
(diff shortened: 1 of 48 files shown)


plugins/upload-assets-to-s3/.toolkitrc.yml
```diff
@@ -1,10 +1,2 @@
 hooks:
   'release:remote': UploadAssetsToS3
-
-options:
-  '@dotcom-tool-kit/upload-static-assets':
-    directory: 'public'
-    bucket: 'ft-next-hashed-assets-prod'
-    destination: 'hashed-assets/uploaded'
-    extensions: 'js,css,map,gz,br,png,jpg,jpeg,gif,webp,svg,ico,json'
-    cacheControl: 'public, max-age=31536000, stale-while-revalidate=60, stale-if-error=3600'
```
jkerr321, line 10: **praise:** nice tidy up getting rid of these unnecessary options

lib/types/src/schema/eslint.ts
```diff
@@ -1,10 +1,10 @@
...
-  config: 'record.unknown?', // @deprecated: use options instead
-  options: 'record.unknown?'
-} as const
-export type ESLintOptions = SchemaOutput<typeof ESLintSchema>
+export const ESLintSchema = z.object({
+  files: z.string().array().default(['**/*.js']),
+  config: z.record(z.unknown()).optional(), // @deprecated: use options instead
+  options: z.record(z.unknown()).optional()
+})
+export type ESLintOptions = z.infer<typeof ESLintSchema>
```
jkerr321, line 8: **thought:** I like the structure we get in the schema here - specifying types as well as defaults and optional values. I don't love the abstraction of the options using `z.infer` because it is additional library code to mentally parse but I understand it and think the descriptive benefits we get in the schema are worth it.

plugins/heroku/src/tasks/production.ts
```diff
@@ -22,23 +22,6 @@ export default class HerokuProduction extends Task<typeof HerokuSchema> {
...
-options:
-  '@dotcom-tool-kit/heroku':
-    scaling:
-      ft-next-static-eu:
-        web:
-          size: standard-1x
-          quantity: 2`
-
-        throw error
-      }
```
jkerr321, line 40: **question:** Do we now lose this level of plugin specific detail in the errors that have been abstracted away from the individual plugin tasks? Why did you make that decision?

lib/types/src/schema.ts
```diff
@@ -1,92 +1,62 @@
...
-  ? unknown
-  : T extends `array.${infer S}`
-  ? S extends SchemaType
-    ? Array<SchemaTypeOutput<S>>
-    : never
-  : T extends `record.${infer S}`
-  ? S extends SchemaType
-    ? Record<string, SchemaTypeOutput<S>>
-    : never
+export type PromptGenerators<T> = T extends z.ZodObject<infer Shape>
```
jkerr321, line 10: **question:** what is `Shape` here?

core/create/src/prompts/options.ts
```diff
@@ -1,126 +1,146 @@
...
 
-  for (const [optionName, optionType] of options) {
-    switch (optionType) {
-      case 'string':
+  for (const { name: optionName, type: optionType, default: optionDefault } of options) {
+    const defaultSuffix = optionDefault
+      ? ` (leave blank to use default value ${styles.code(JSON.stringify(optionDefault))})`
+      : ''
+
+    switch (optionType._def.typeName as z.ZodFirstPartyTypeKind) {
```
jkerr321, line 33: **suggestion:** before this change this line was `switch (optionType)`, and now it is `switch (optionType._def.typeName as z.ZodFirstPartyTypeKind)` which is less readable and understandable. Could this be made easier to read with descriptive variable names (or some other solution)?

jkerr321:
**thought:** I like the descriptiveness and structure that zod adds to the schema files in Tool Kit, however elsewhere I find the use of zod confusing. It is another abstraction on top of TS that (in my opinion), outside of the schema files, makes the code less easy to mentally parse for developers. That said, the place where it is most confusing for me is in some of the cli files which I found confusing to begin with 😬 I am interested in hearing what other people think!

core/create/src/prompts/options.ts
ivomurrell: I'll split it out into a variable assignment and add a comment explaining why we're using the `_def` field – does that sound better?

plugins/heroku/src/tasks/production.ts
ivomurrell: These errors will now never be hit because we do option validation when loading the plugins rather than having to check the options in the `run()` method. Previously you had to remember that even though the fields in the schema weren't optional – and therefore the given option must be specified for your Tool Kit configuration to be valid – we didn't actually do any checks when loading your configuration. So even though the options type could say an option field like `importantId` was of type `string`, as opposed to `string | undefined`, it could actually still be undefined if a user passed invalid config. So it was prone to semantic errors if you forgot to write specific checks.

As to lacking specific detail, using zod we can produce an error like
```
Options are defined in your Tool Kit configuration that are the wrong types:

- @dotcom-tool-kit/next-router has the issue(s): Expected string, received number at "appName"
- @dotcom-tool-kit/example has the issue(s): Required at "importantId"

Please update the options so that they are the expected types.
```
Do you think this is sufficient for telling users what an issue is and how they can resolve it?

lib/types/src/schema/eslint.ts
ivomurrell: Imo it isn't any worse than the `SchemaOutput` type we were using previously to extract the `*Options` types. And helpfully in both cases, VSCode will also show you what the resolved type is if you hover over `ESLintOptions`!

lib/types/src/schema.ts
ivomurrell: `Shape` is the name of a type we're declaring using the `infer` keyword. It's a way to extract a type parameter from a generic type like `Array<string>`. You could use `Array<infer Item>` to get an `Item` type, though it could equally be called the `Value` type or the `Ivo` type. The `infer` keyword is useable in these conditional types (i.e., types that looks like `T extends Foo<infer Bar> ? U : V`) which you can read more about in the [TypeScript handbook](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#inferring-within-conditional-types).

In this case `Shape` will be the first type parameter of the `z.ZodObject` type. This type parameter is an object whose values are the zod schemas of each value in the object's schema. In practical terms it's the type of the object you pass to `z.object` in a schema definition. So if you declared a schema like
```ts
const example: z.ZodObject = z.object({
  configPath: z.string(),
  id: z.number()
})
```
The inferred `Shape` type (i.e., `ZodObject`'s first type parameter) would be the type of
```ts
{
  configPath: z.string(),
  id: z.number()
}
```
so something like
```ts
{
  configPath: z.ZodString,
  id: z.ZodNumber
}
```
In `PromptGenerators` we then pass each of `Shape`'s values (the zod schemas for each option) to the `SchemaPromptGenerator` generic type.

lib/types/src/schema.ts
```diff
@@ -1,92 +1,62 @@
...
-  '@dotcom-tool-kit/next-router'?: NextRouterOptions
-  '@dotcom-tool-kit/prettier'?: PrettierOptions
-  '@dotcom-tool-kit/lint-staged-npm'?: LintStagedNpmOptions
-  '@dotcom-tool-kit/babel'?: BabelOptions
-  '@dotcom-tool-kit/circleci'?: CircleCIOptions
-  '@dotcom-tool-kit/cypress'?: CypressOptions
-  '@dotcom-tool-kit/typescript'?: TypeScriptOptions
+  [plugin in keyof typeof Schemas]?: typeof Schemas[plugin] extends z.ZodTypeAny
+    ? z.infer<typeof Schemas[plugin]>
+    : never
```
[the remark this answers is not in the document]
ivomurrell: Not sure if there's a more verbose way to write it but I'll write a comment explaining what the transform is at least.

core/create/src/prompts/options.ts
jkerr321: Yep!

plugins/heroku/src/tasks/production.ts
jkerr321: Ok I'm sold! Out of interest... how would the Heroku scaling error look as that's a funky nested object? the previous error was quite descriptive about what was required and I'm interested how Zod handles that

lib/types/src/schema.ts
jkerr321: 🙏 Thank you for the detailed explanation

plugins/heroku/src/tasks/production.ts
ivomurrell: Some example Heroku errors with zod:
* if you pass an empty object:
```
@dotcom-tool-kit/heroku has the issue(s): Required at "pipeline"; Required at "systemCode"; Required at "scaling"
```
* when you forget to include the process type:
```
@dotcom-tool-kit/heroku has the issue(s): Expected object, received number at "scaling.sandbox-eu.size"; Expected object, received number at "scaling.sandbox-eu.quantity"
```
* when you mix up the size and quantity fields:
```
@dotcom-tool-kit/heroku has the issue(s): Expected string, received number at "scaling.sandbox-eu.big.size"
```

And whilst we checked for the `pipeline` and `systemCode` fields before and gave nice errors before, we [didn't actually validate](https://github.com/Financial-Times/dotcom-tool-kit/blob/18e4a271c43711a1160faed4662f41b4c5329880/plugins/heroku/src/tasks/production.ts#L42-L50) the `scaling` object at all and assumed users followed the documentation exactly, which would result in confusing JavaScript's `TypeError`s!

jkerr321: Thanks for getting the additional error info. These errors are not as readable as the previous error - if you were new to Tool Kit I think you would immediately know what to change if you saw the old error, but you wouldn't in some cases with the new error (I'm particularly thinking about the nested object nature of the scaling option).

But I also accept that the old error didn't cover every case and this Heroku option is an outlier in terms of its complexity so the new error is more useful in most cases. My suggestion is a litttle bit more direction for the user in the error with an additional line something like "please refer to the readme for this plugin for the full config required for this plugin". (Again I see pros and cons to that - the most definitive place to refer a user to is the schema file itself but I think the readme is better in terms of readability, even if we are having to assume that it will be kept up to date with the schema). What do you think? 😬 😁

jkerr321 (approved):
Nice! Thank you for the additional comments and clarifications 👍

PR #355, opened 2023-01-20
CPP-1070 Uploadassetstos3 improvements

jkerr321:
# Description

This PR improves some aspects of the `UploadAssetsToS3` task. Each change is captured in a commit. The final change is the one that I'd like consideration on:

### Logging when files not found
Currently if no files match the provided directory and glob the code continues to run and no logging indicates to the user that files haven't been found and uploaded. This change throws an error if no files are found that match the provided glob.

### Documentation changes for multiple buckets
Update documentation and default options to reflect the following points:
- My understanding is that support for multiple production buckets in this task is for the case where there exists a `prod` and `prod-us` bucket. However, the common use case (I haven't yet found a conflicting example) in Customer Products for the projects where there is a `prod` and `prod-us` S3 bucket is that the `prod-us` bucket is pre configured in AWS to replicate the `prod` bucket. I.e. file changes should only be uploaded the the `prod` bucket then AWS will do the rest to replicate them to `prod-us`.
- If a user does want to upload to two buckets, the `UploadAssetsToS3` task only allows for one set of credentials to be passed as options so the credentials to access both buckets need to be the same

### Better error handling with multiple buckets

In the existing version of the `UploadAssetsToS3` task there is a scenario where [the following block of code](https://github.com/Financial-Times/dotcom-tool-kit/blob/main/plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts#L56-L69) does not return an error even when it has not executed successfully:

```ts
        for (const bucket of bucketByEnv) {
          const params = {
            Bucket: bucket,
            Key: key,
            Body: body,
            ACL: 'public-read',
            ContentType: `${type}; charset=utf-8`,
            ContentEncoding: encoding,
            CacheControl: options.cacheControl
          }
          currentBucket = params.Bucket
          const data = await s3.upload(params).promise()
          this.logger.info(`Uploaded ${styles.filepath(filepath)} to ${styles.URL(data.Location)}`)
        }
```

The failure scenario is:
- Two buckets are specified for upload (e.g. `['ft-next-service-registry-prod', 'ft-next-service-registry-prod-us']`)
     AND
- Uploads are successful to the first bucket
    AND
- There is an error in uploads to the second bucket

In this scenario
- Success logs are output for all files uploaded in the first bucket (e.g. `Uploaded public/user-facing-services.json to https://ft-next-service-registry-prod.s3.amazonaws.com/test/user-facing-services.json`)
- No error is thrown
- The task finishes running and exits with no further logging

This issue is caused by v2 of the aws-sdk not properly supporting native promises and async await causing unexpected behaviour.

One solution is to wrap all the promises in a `Promise.all` and then handle the successful responses once they've all resolved. That was my initial fix and is still captured in [a90fd11](https://github.com/Financial-Times/dotcom-tool-kit/pull/355/commits/a90fd118def9ad883073b491937c845c2f26af31).

A better solution is to migrate to v3 of the aws-sdk which properly implements support for async await in all cases.

Below is the messaging output by `UploadAssetsToS3` before and after this change, where the upload to the `['ft-next-service-registry-prod-us']` rejects with an error.

|  `prodBucket` config value in `toolkitrc.yml` | Existing | Expected behaviour? | After this PR |
|---------|----------------------|-|-------------------------|
| `['ft-next-service-registry-prod']`| image |Yes| image |
| `['ft-next-service-registry-prod-us']`| image | Yes| image|
| `['ft-next-service-registry-prod', 'ft-next-service-registry-prod-us']`| image |  No | image |

# Checklist:

### DIFF
--- a/plugins/upload-assets-to-s3/README.md
+++ b/plugins/upload-assets-to-s3/README.md
@@ -28,6 +28,7 @@ plugins:
 | `directory` | the folder in the project whose contents will be uploaded to S3 | 'public' |
 | `reviewBucket` | the development or test S3 bucket | `['ft-next-hashed-assets-preview']` |
 | `prodBucket` | production S3 bucket/s; an array of strings. The same files will be uploaded to each. _Note: most Customer Products buckets that have a `prod` and `prod-us` version are already configured in AWS to replicate file changes from one to the other so you don't need to specify both here. Also, if multiple buckets are specified the same credentials will need to be valid for both for the upload to be successful_ | `['ft-next-hashed-assets-prod']` |
+| `region` | the AWS region your buckets are stored in (let the Platforms team know if you need to upload to multiple buckets in multiple regions) | eu-west-1 |
 | `destination` | the destination folder for uploaded assets. Set to `''` to upload assets to the top level of the bucket | 'hashed-assets/page-kit' |
 | `extensions` | file extensions to be uploaded to S3 | 'js,css,map,gz,br,png,jpg,jpeg,gif,webp,svg,ico,json' |
 | `cacheControl` | header that controls how long your files stay in a CloudFront cache before CloudFront forwards another request to your origin | 'public, max-age=31536000, stale-while-revalidate=60, stale-if-error=3600' |
--- a/plugins/upload-assets-to-s3/README.md
+++ b/plugins/upload-assets-to-s3/README.md
@@ -21,7 +21,7 @@ plugins:
 ## Options
 | Key | Description | Default value |
 |-|-|-|
-| `accessKeyIdEnvVar` | variable name of the project's aws access key id. If uploading to multiple buckets the same credentials will need to work for both. | no default value - for backwards compatability the plugin falls back to the default value for `accessKeyId` |
+| `accessKeyIdEnvVar` | variable name of the project's aws access key id. If uploading to multiple buckets the same credentials will need to work for all. | no default value - for backwards compatability the plugin falls back to the default value for `accessKeyId` |
 | `secretAccessKeyEnvVar` | variable name of the project's aws secret access key | no default value - for backwards compatability the plugin falls back to the default value for `secretAccessKey` |
 | `accessKeyId` | **DEPRECATED** variable name of the project's aws access key id | 'aws_access_hashed_assets' |
 | `secretAccessKey` | **DEPRECATED** variable name of the project's aws secret access key | 'aws_secret_hashed_assets' |
--- a/plugins/upload-assets-to-s3/README.md
+++ b/plugins/upload-assets-to-s3/README.md
@@ -21,13 +21,13 @@ plugins:
 ## Options
 | Key | Description | Default value |
 |-|-|-|
-| `accessKeyIdEnvVar` | variable name of the project's aws access key id | no default value - for backwards compatability the plugin falls back to the default value for `accessKeyId` |
+| `accessKeyIdEnvVar` | variable name of the project's aws access key id. If uploading to multiple buckets the same credentials will need to work for both. | no default value - for backwards compatability the plugin falls back to the default value for `accessKeyId` |
 | `secretAccessKeyEnvVar` | variable name of the project's aws secret access key | no default value - for backwards compatability the plugin falls back to the default value for `secretAccessKey` |
 | `accessKeyId` | **DEPRECATED** variable name of the project's aws access key id | 'aws_access_hashed_assets' |
 | `secretAccessKey` | **DEPRECATED** variable name of the project's aws secret access key | 'aws_secret_hashed_assets' |
 | `directory` | the folder in the project whose contents will be uploaded to S3 | 'public' |
-| `reviewBucket` | the development or test S3 bucket | ['ft-next-hashed-assets-preview'] |
-| `prodBucket` | production S3 bucket/s; an array of strings. The same files will be uploaded to each | ['ft-next-hashed-assets-prod', 'ft-next-hashed-assets-prod-us'] |
+| `reviewBucket` | the development or test S3 bucket | `['ft-next-hashed-assets-preview']` |
+| `prodBucket` | production S3 bucket/s; an array of strings. The same files will be uploaded to each. Note: most Customer Products buckets that have a `prod` and `prod-us` version are already configured in AWS to replicate file changes from one to the other so you don't need to specify both here. Also, if multiple buckets are specified the same credentials will need to be valid for both for the upload to be successful | `['ft-next-hashed-assets-prod']` |
 | `destination` | the destination folder for uploaded assets. Set to `''` to upload assets to the top level of the bucket | 'hashed-assets/page-kit' |
 | `extensions` | file extensions to be uploaded to S3 | 'js,css,map,gz,br,png,jpg,jpeg,gif,webp,svg,ico,json' |
 | `cacheControl` | header that controls how long your files stay in a CloudFront cache before CloudFront forwards another request to your origin | 'public, max-age=31536000, stale-while-revalidate=60, stale-if-error=3600' |
@@ -38,7 +38,7 @@ Example:
   '@dotcom-tool-kit/upload-assets-to-s3':
     accessKeyId: AWS_ACCESS
     secretAccessKey: AWS_KEY
-    prodBucket: ['ft-next-service-registry-prod', 'ft-next-service-registry-prod-us']
+    prodBucket: ['ft-next-service-registry-prod']
     reviewBucket: ['ft-next-service-registry-dev']
     destination: ''
 ```
--- a/plugins/upload-assets-to-s3/README.md
+++ b/plugins/upload-assets-to-s3/README.md
@@ -27,7 +27,7 @@ plugins:
 | `secretAccessKey` | **DEPRECATED** variable name of the project's aws secret access key | 'aws_secret_hashed_assets' |
 | `directory` | the folder in the project whose contents will be uploaded to S3 | 'public' |
 | `reviewBucket` | the development or test S3 bucket | `['ft-next-hashed-assets-preview']` |
-| `prodBucket` | production S3 bucket/s; an array of strings. The same files will be uploaded to each. Note: most Customer Products buckets that have a `prod` and `prod-us` version are already configured in AWS to replicate file changes from one to the other so you don't need to specify both here. Also, if multiple buckets are specified the same credentials will need to be valid for both for the upload to be successful | `['ft-next-hashed-assets-prod']` |
+| `prodBucket` | production S3 bucket/s; an array of strings. The same files will be uploaded to each. _Note: most Customer Products buckets that have a `prod` and `prod-us` version are already configured in AWS to replicate file changes from one to the other so you don't need to specify both here. Also, if multiple buckets are specified the same credentials will need to be valid for both for the upload to be successful_ | `['ft-next-hashed-assets-prod']` |
 | `destination` | the destination folder for uploaded assets. Set to `''` to upload assets to the top level of the bucket | 'hashed-assets/page-kit' |
 | `extensions` | file extensions to be uploaded to S3 | 'js,css,map,gz,br,png,jpg,jpeg,gif,webp,svg,ico,json' |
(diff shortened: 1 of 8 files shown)


plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
```diff
@@ -53,6 +53,7 @@ export default class UploadAssetsToS3 extends Task<typeof UploadAssetsToS3Schema
         const data = await s3.upload(params).promise()
         this.logger.info(`Uploaded ${styles.filepath(filepath)} to ${styles.URL(data.Location)}`)
       } else {
+        const uploadPromises: Array<Promise<any>> = []
```
nickramsbottom, line 56: **suggestion**: is it possible to type the `any` here given you are pushing a specific promise to the array on line 69?

plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
```diff
@@ -64,9 +65,16 @@ export default class UploadAssetsToS3 extends Task<typeof UploadAssetsToS3Schema
             CacheControl: options.cacheControl
           }
           currentBucket = params.Bucket
-          const data = await s3.upload(params).promise()
-          this.logger.info(`Uploaded ${styles.filepath(filepath)} to ${styles.URL(data.Location)}`)
+          const data = s3.upload(params).promise()
+          uploadPromises.push(data)
         }
+        await Promise.all(uploadPromises).then((resolvedUploadPromises) => {
```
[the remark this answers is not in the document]
jkerr321: Interesting... we do want to throw and exit the process if any of them fail as this task is used in deployment (on CircleCI) and we don't want partial upload to equal success. We could use all `allSettled` though and throw once all the promises are returned with a collated list of the failures.

plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
jkerr321: Yes! I have a question about this as it's my first time writing TS (exciting!). I think it would be  `Array<Promise<aws.S3.ManagedUpload.SendData>>>` but I wasn't sure if it should be more generic than that?

nickramsbottom: I'm pretty sure you can steal the return type off of `s3.upload(params).promise` somehow. I hadn't thought about this before though, how typing the failure as well as the success works

plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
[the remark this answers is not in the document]
ivomurrell: either is probably fine, it's a case of whether we want to fail fast or not if something goes wrong. i guess we'd rather have the rest of the assets uploaded even if one upload fails rather than leaving the assets partially updated? in which case `allSettled` would be the choice.

plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
nickramsbottom: ahhh this is vaguely coming back to me: https://stackoverflow.com/questions/50071115/typescript-promise-rejection-type

I think you check the type of reject before you act upon it but you can't type it

plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
```diff
@@ -64,9 +65,16 @@ export default class UploadAssetsToS3 extends Task<typeof UploadAssetsToS3Schema
             CacheControl: options.cacheControl
           }
           currentBucket = params.Bucket
-          const data = await s3.upload(params).promise()
-          this.logger.info(`Uploaded ${styles.filepath(filepath)} to ${styles.URL(data.Location)}`)
+          const data = s3.upload(params).promise()
+          uploadPromises.push(data)
```
ivomurrell, line 69: **suggestion**: we could rewrite this as a [`.map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) over `bucketByEnv` rather than a for loop, which will save you creating an empty array and pushing to it each iteration cycle. something like
```ts
bucketByEnv.map((bucket) => {
  const params = {
     ... // snip
  }
  currentBucket = params.Bucket
  return s3.upload(params).promise()
})

File: plugins/upload-assets-to-s3/README.md [hunk 4]
```diff
@@ -21,13 +21,13 @@ plugins:
 ## Options
 | Key | Description | Default value |
 |-|-|-|
-| `accessKeyIdEnvVar` | variable name of the project's aws access key id | no default value - for backwards compatability the plugin falls back to the default value for `accessKeyId` |
+| `accessKeyIdEnvVar` | variable name of the project's aws access key id. If uploading to multiple buckets the same credentials will need to work for both. | no default value - for backwards compatability the plugin falls back to the default value for `accessKeyId` |
```
ivomurrell, line 24: **nitpick**: doesn't have to be just two either!

```suggestion
| `accessKeyIdEnvVar` | variable name of the project's aws access key id. If uploading to multiple buckets the same credentials will need to work for all. | no default value - for backwards compatability the plugin falls back to the default value for `accessKeyId` |
```

plugins/upload-assets-to-s3/README.md
```diff
@@ -21,13 +21,13 @@ plugins:
...
 | `accessKeyId` | **DEPRECATED** variable name of the project's aws access key id | 'aws_access_hashed_assets' |
 | `secretAccessKey` | **DEPRECATED** variable name of the project's aws secret access key | 'aws_secret_hashed_assets' |
 | `directory` | the folder in the project whose contents will be uploaded to S3 | 'public' |
-| `reviewBucket` | the development or test S3 bucket | ['ft-next-hashed-assets-preview'] |
-| `prodBucket` | production S3 bucket/s; an array of strings. The same files will be uploaded to each | ['ft-next-hashed-assets-prod', 'ft-next-hashed-assets-prod-us'] |
+| `reviewBucket` | the development or test S3 bucket | `['ft-next-hashed-assets-preview']` |
+| `prodBucket` | production S3 bucket/s; an array of strings. The same files will be uploaded to each. Note: most Customer Products buckets that have a `prod` and `prod-us` version are already configured in AWS to replicate file changes from one to the other so you don't need to specify both here. Also, if multiple buckets are specified the same credentials will need to be valid for both for the upload to be successful | `['ft-next-hashed-assets-prod']` |
```
ivomurrell, line 30: **suggestion(non-blocking)**: maybe we could put the note in italics so that it's easier for people to differentiate it when they're skimming the table?

```suggestion
| `prodBucket` | production S3 bucket/s; an array of strings. The same files will be uploaded to each. _Note: most Customer Products buckets that have a `prod` and `prod-us` version are already configured in AWS to replicate file changes from one to the other so you don't need to specify both here. Also, if multiple buckets are specified the same credentials will need to be valid for both for the upload to be successful_ | `['ft-next-hashed-assets-prod']` |
```

ivomurrell (commented):
**praise**: this is one of the best PR descriptions I've seen for a while: very thorough, well-formatted, and well explained 😁

> I think this issue is caused due to with looping with promises - a race condition where resolved and rejected promises are returning unpredictably and interfering with each other.

**issue**: This shouldn't be what's happening: the `await` will effectively pause (or 'yield') the code until the promise has either resolved or rejected, so the `await`'s in the for loop will be happening one after the other – the first yielding until it's resolved before the second promise is created and subsequently rejects – never simultaneously. So I don't understand what was going wrong and I'm nervous we're missing some other logical error which this change temporarily alleviates rather than fixing. Would we be able to timebox pairing together on the issue tomorrow and if no progress is made we can merge this change which seems to fix the issue even if we don't understand why?

plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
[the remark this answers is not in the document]
jkerr321: When I change the code to `Promise.allSettled` I get the error 
```
Property 'allSettled' does not exist on type 'PromiseConstructor'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2020' or later.ts(2550)
```
I actually think it's fine with `all` - updating it to `allSettled` makes the code slightly more difficult to read and doesn't add lots in terms of functionality. For the use case that I'm aware of where this task is used for deployment, the deployment will either pass or fail, and in the latter case will need to be fixed (in theory!) regardless of how many files successfully upload. 

Either way I have added a card to add support for ES2020 as this will not be the last time ES2020 features will be useful to us: https://financialtimes.atlassian.net/browse/CPP-1248

plugins/upload-assets-to-s3/test/tasks/upload-assets-to-s3.test.ts
```diff
@@ -51,10 +52,10 @@ describe('upload-assets-to-s3', () => {
...
 
     await task.run()
-
-    const s3 = mocked(mockedAWS.S3.mock.instances[0])
-    expect(s3.upload).toHaveBeenCalledTimes(2)
-    expect(s3.upload.mock.calls[0][0]).toHaveProperty('Key', 'testdir/nested/test.js.gz')
+    const s3 = mockedS3Client.mock.instances[0]
+    const putObjectCommand = mockedPutObjectCommand.mock.instances[0]
+    expect(s3.send).toHaveBeenCalledTimes(1)
+    // expect(putObjectCommand.mock.calls[0][0]).toHaveProperty('Key', 'testdir/nested/test.js.gz')
```
jkerr321, line 58: @ivomurrell did you get to the bottom of how to test this?

ivomurrell: I forgot 🙊 I'll do that today

ivomurrell:
@nickramsbottom pointed it out – and I think it's worth noting – the `package-lock.json` looks to have had significant changes in this PR... it has 17k insertions and 14k deletions according to GitHub! it looks like we've deleted the lockfile and installed over it, ignoring any previous package versions and using the newest available. but that's not actually true: we've updated to `aws-sdk@3` and that has caused the apparent rehaul of the lockfile. but it's also not because `aws-sdk@3` completely changes how all our dependencies are resolved either! it's actually because of git's diffing algorithm...

`aws-sdk@3` modularises their `aws-sdk` into many, many library packages, which are now installed as separate dependencies rather than one monolithic dependency. this means that there are many new additions to the lockfile, and right near the top thanks to the `@aws-sdk` namespace being alphabetically sorted. this throws off git's diffing algorithm and it's no longer able to find similarities between the two lockfiles as the line offsets between the original place for a dependency and its new place in the lockfile are out by thousands of lines, and the algorithm's heuristics cause it to break off the search prematurely (or something along those lines, I'm sure it's much more complicated in reality...) you can see this by skimming the diff in your terminal and noting that each block of deletions and insertions are comparing packages multiple letters apart in the alphabet.

now, if you remove the heuristics git uses to speed up its diffing by specifying the `minimal` algorithm instead ([git docs](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---diff-algorithmpatienceminimalhistogrammyers), [article comparing the algorithms](https://luppeng.wordpress.com/2020/10/10/when-to-use-each-of-the-git-diff-algorithms/)) you can see the real difference:
```bash
$ git diff --diff-algorithm=minimal --shortstat main origin/uploadassetstos3-improvements -- package-lock.json
1 file changed, 3494 insertions(+), 276 deletions(-)
```
only (!) 3k insertions and 276 deletions.

something to try out on lockfiles in the future if you're concerned they've been rebuild somehow!

tsconfig.settings.json
```diff
@@ -8,7 +8,7 @@
     "strict": true,
     "target": "ES2019",
     "types": ["node"],
-    "lib": ["ES2019", "dom"],
```
jkerr321: **praise**: nice improvement, thank you!

ivomurrell (approved):
@jkerr321 has said she is happy with my changes, and I'm happy with hers, so let's get this collaborative PR merged!

PR #356, opened 2023-01-23
CPP-1049 Split Heroku tasks out from plugins that define hooks

ivomurrell:
# Description

We currently only support Heroku as a method for deploying projects via Tool Kit, which has meant we’ve made a few assumptions about how users will consume the Tool Kit infrastructure needed to deploy code. In particular, all the hooks to install your deploy code into CircleCI are defined in a plugin called `circleci-heroku`. These hooks are actually already designed to be generic – they just install calls to the Tool Kit orb into your CircleCI config – so no logic needs to change. However, the name of the plugin is confusing now that we want to allow other deployment methods, such as Serverless, to use the hooks, not just Heroku.

I've now renamed the plugin to the generic `circleci-deploy`. This plugin no longer imports the `heroku` plugin, which was only used to declare default task associations for the hooks. Now, a user who wants to deploy to Heroku should import both the `circleci-deploy` and `heroku` apps and specify these associations themselves. This seems like a fair bit of faff for such a common use case, and that's why we've told people to use `backend-app` to pull in all these plugins and set up the default hooks in the past. However, this plugin is also not suitably generic, as we want to support people who have a backend app which deploys to Serverless in the future too. So, in addition, `backend-app` has been renamed to `backend-heroku-app` and pulls in the new `circleci-deploy` plugin and `heroku` plugin. In the future we can add a `serverless` plugin and a `backend-serverless-app` plugin to define the tasks and config required to deploy to Serverless and utilise the generic `circleci-deploy` plugin.

To maintain backwards compatibility I've kept the old plugins, `circleci-heroku` and `backend-app`. These plugins now re-export the new `circleci-deploy` and `backend-heroku-app` plugins whilst retaining the same behaviour as they had previously. They have been marked as deprecated and users are pointed towards the new plugins. We will hopefully be able to remove the plugins entirely in the future.

# Checklist:


plugins/backend-app/readme.md
```diff
@@ -1,26 +1,5 @@
...
-npm install --save-dev @dotcom-tool-kit/backend-app
-```
-
-And add it to your repo's `.toolkitrc.yml`:
-
-```yaml
-plugins:
-    - '@dotcom-tool-kit/backend-app'
-```
+This plugin is deprecated. It is kept to preserve backwards compatibility for projects that previously depended on it, but just imports the `backend-heroku-app`, which supersedes this plugin.
```
apaleslimghost, line 5: **suggestion**: could the deprecated plugins print a warning on `--install`? they could include a dummy hook that just has:

```js
class DeprecatedHook extends Hook {
	check() {
		this.logger.warn(deprecationMessage)
		return true
	}
}
```

would be useful in the future if we decide to deprecate any other plugins

apaleslimghost: **question**: should these plugins be marked as deprecated on npm too?

plugins/frontend-app/.toolkitrc.yml
```diff
@@ -1,6 +1,6 @@
 plugins:
```
apaleslimghost, line 1: **question**: do we want to rename this plugin too? right now it would be weird to have a frontend app not hosted on Heroku, but i could see that happening in the future with e.g. Compute@Edge

apaleslimghost (approved):
looking great, a few non-blocking comments

plugins/backend-app/readme.md
ivomurrell: I think we can restrict this to just rely on npm deprecation warnings and documentation as I've found trying to overload the hook installation logic with the logging can be a little awkward (some deprecated plugins don't have any source code atm so would need to add TypeScript config or write OOP in JS; we would need to set unique names for each of the hooks which would then show up in `--help` too; the `check` method runs on every run, not just on `--install`, so there logging would be pretty noisy.) I think the other places highlighting deprecation should be sufficient?

ivomurrell: I'll add the deprecation messages after this PR is merged!

PR #358, opened 2023-02-03
CPP:1051 - feat: add serverless plugin

jkerr321:
Create a serverless plugin for managing Lambdas. This PR creates the initial plugin structure as well as a `ServerlessRun` task for use on the `run:local` hook.

The `ServerlessRun` task uses the `serverless offline start` command which mirrors how most CP lambdas are run locally. See [the Supporting AWS lambdas in Tool Kit](https://financialtimes.atlassian.net/wiki/spaces/CPP/pages/7966326832/Supporting+AWS+lambdas+in+Tool+Kit ) confluence page for the initial discovery around how this plugin would be implemented to best work with existing CP Lambda practices.

# Checklist:

### DIFF
--- a/plugins/serverless/src/tasks/run.ts
+++ b/plugins/serverless/src/tasks/run.ts
@@ -1,0 +1,62 @@
+import { Task } from '@dotcom-tool-kit/types'
+import { ServerlessOptions, ServerlessSchema } from '@dotcom-tool-kit/types/lib/schema/serverless'
+import { spawn } from 'child_process'
+import { VaultEnvVars } from '@dotcom-tool-kit/vault'
+import { hookConsole, hookFork } from '@dotcom-tool-kit/logger'
+import getPort from 'get-port'
+import waitPort from 'wait-port'
+
+export default class ServerlessRun extends Task<typeof ServerlessSchema> {
+  static description = ''
+
+  static defaultOptions: ServerlessOptions = {
+    useVault: true,
+    ports: [3001, 3002, 3003]
+  }
+
+  async run(): Promise<void> {
+    const { useVault, ports, configPath } = this.options
+
+    let vaultEnv = {}
+
+    if (useVault) {
+      const vault = new VaultEnvVars(this.logger, {
+        environment: 'development'
+      })
+
+      vaultEnv = await vault.get()
+    }
+
+    const port =
+      Number(process.env.PORT) ||
+      (await getPort({
+        port: ports
+      }))
+
+    this.logger.verbose('starting the child serverless process...')
+    const args = ['offline', 'start', '--host', 'local.ft.com', '--httpPort', `${port}`]
+    if (configPath) {
+      args.push('--config', './serverless.yml')
+    }
+
+    const child = spawn('serverless', args, {
+      env: {
+        ...vaultEnv,
+        PORT: port.toString(),
+        ...process.env
+      }
+    })
+
+    hookFork(this.logger, 'serverless', child)
+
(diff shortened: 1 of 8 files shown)


plugins/serverless/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "../../tsconfig.settings.json",
+  "compilerOptions": {
+    "outDir": "lib",
+    "rootDir": "src"
+  },
+  "references": [
+    {
+      "path": "../../lib/types"
+    }
```
ivomurrell, line 10: **nitpick**: the other Tool Kit packages you use in `serverless` should be included in the `references` object too, e.g., `vault`, `logger`, etc. this ensures that `serverless` is compiled after all of them and means you won't get build errors from trying to import modules that haven't been built yet (though that usually only will happen with a clean build).

jkerr321: Ahh interesting, I didn't know that, thanks for the explanation!

jkerr321: I've updated that now and I have another question :D - the [Node task includes the options lib](https://github.com/Financial-Times/dotcom-tool-kit/blob/main/plugins/node/tsconfig.json#L18) in its list of references, but I don't see anywhere in the Node plugin where it's directly required - is that just a mistake or am I missing something that means I should potentially be adding other references to my `tsconfig.json` that aren't directly required?

(Also, I see that the `state` lib is also not included in the list of refs for the Node plugin, and that is directly required. I assume that is a mistake, I can make a little PR to add that in...)

ivomurrell: yep, that is a mistake! we're not very good at maintaining those because usually it doesn't present a problem. it will only cause the first `tsc` build of the project to fail if they're misconfigured and any subsequent runs should work just fine so it's not really a priority. but it is 1) more _correct_ and 2) presents a confusing experience and a bad first impression of the monorepo if your first build fails!

ivomurrell: I would prefer a hard error if `references` is configured incorrectly tbh but it doesn't look like such an option exists atm. `references` [seems](https://www.typescriptlang.org/docs/handbook/project-references.html#what-is-a-project-reference) like it makes a few more complicated and subtle changes to your project anyway so maybe that's why it's not strictly enforced.

plugins/serverless/src/tasks/run.ts
```diff
@@ -0,0 +1,62 @@
+import { Task } from '@dotcom-tool-kit/types'
+import { ServerlessOptions, ServerlessSchema } from '@dotcom-tool-kit/types/lib/schema/serverless'
+import { spawn } from 'child_process'
+import { VaultEnvVars } from '@dotcom-tool-kit/vault'
+import { hookConsole, hookFork } from '@dotcom-tool-kit/logger'
+import getPort from 'get-port'
+import waitPort from 'wait-port'
+
+export default class ServerlessRun extends Task<typeof ServerlessSchema> {
+  static description = ''
```
ivomurrell, line 10: **nitpick**: could you add a short description for the task here? we should get into the habit of doing this, it will enable us to display the tasks and hooks to users in a friendlier way in the future.

ivomurrell (approved):
an exciting start to the plugin!

plugins/serverless/src/tasks/run.ts
jkerr321: Oh yes good point!

jkerr321: Done :)

PR #359, opened 2023-02-08
CPP-1259 Revamp CircleCI plugin to allow much more flexible CircleCI configuration

ivomurrell:
# Description

When first writing the `circleci` plugin's `.install()` behaviour, I found that defining the shape of a CircleCI config file could be quite complicated, but that there were some general patterns which we could abstract into something simpler. We initially only allowed hooks implementing the CircleCI installation code to specify a single job, but soon found this to be too limiting. Some hooks require additional options on other jobs, or need to specify options outside of a workflow, or need to declare a new job etc. We kept piling on more abstractions and increasing the code complexity to the point where it was fairly unmaintainable. Now, we need to support a hook that can install multiple jobs so that OIDC authentication can work for the Serverless plugin, and rather than adding another pile of straw, I've resolved to rewrite the architecture instead.

Now, each hook specifies a partial view of the CircleCI configuration, declaring the parts the hook is concerned with, e.g., a new job in the `tool-kit` and `nightly` workflows. These partial views are then merged together on top of a default configuration. However, each partial view needs to define the whole nested configuration structure preceding the changes it's concerned with (the path to a new job is `workflows["tool-kit"].jobs[].[jobName]` for instance), and there is a lot of repetitive boilerplate each job needs but hooks needn't be concerned with. Therefore, a generator function, `generateConfigWithJob`, is available that allows you to specify a job's parameters and get the partial view in return. It
is then relatively easy to make further arbitrary modifications to this partial view via object spreads if required.

This new approach allows for any hook to make arbitrary changes to the CircleCI config that don't require any new code support in the `circleci` plugin. I've also taken the chance to remove the '`postinstall`' logic which would previously make further modifications to the config object after all hooks had finished specifying their jobs etc. This would, before these changes, set up tag filters if tagged releases were used, and use a custom executor if we wanted to override the Docker image or tag. I've found that both these tasks can just be done unconditionally regardless of your Tool Kit options and have consequently removed the '`postinstall`' logic that checks if certain hooks have been used. This gives a couple of benefits: 1) it makes the code much simpler, and 2) it reduces the number of assumptions we have to make about the structure of the computed config when running `.commitInstall()`. It does result in configuration files that are more verbose and now contain some settings that may be redundant but I think the benefits outweigh the costs in this case.

# Checklist:


jkerr321:
Yay! Does the way a user interacts with the CircleCi plugin change as a result of this PR?

ivomurrell:
> Yay! Does the way a user interacts with the CircleCi plugin change as a result of this PR?

No! The only user-facing change this PR will make is updating a user's CircleCI config to the latest settings. The rest is all under the hood and focuses on improving the ability of developers – both Tool Kit maintainers or devs making custom plugins – to define modifications to CircleCI configuration in whatever which way they need via hooks.

plugins/circleci/src/index.ts
```diff
@@ -1,16 +1,18 @@
...
+  config = generateConfigWithJob({ name: this.job, requires: ['tool-kit/setup'], addToNightly: true })
 }
 
 export class TestCI extends CircleCiConfigHook {
   job = 'tool-kit/test'
-  jobOptions = { requires: [new BuildCI(this.logger).job] }
-  addToNightly = true
+  config = generateConfigWithJob({
+    name: this.job,
+    requires: [new BuildCI(this.logger).job],
```
jkerr321, line 13: is `this.logger` the logger defined in the abstract `Hook` class [here](https://github.com/Financial-Times/dotcom-tool-kit/blob/main/lib/types/src/index.ts#L173)? 

If so, (and even if not so 😁 ) why does it need to be passed into `new BuildCI`?

plugins/circleci/src/circleci-config.ts
```diff
@@ -1,29 +1,178 @@
...
+import merge from 'lodash/merge'
+import mergeWith from 'lodash/mergeWith'
+import path from 'path'
+import type { PartialDeep } from 'type-fest'
+import YAML from 'yaml'
+
+const majorOrbVersion = '3'
 
-const majorOrbVersion = '2'
+export type CircleCIState = CircleConfig
```
jkerr321, line 18: **question:** why are you creating a new type here? Why not use just `CircleConfig`? Because `CircleCIState` changes as the config functions are run?

plugins/circleci/src/index.ts
ivomurrell: yes it is, and it needs to be passed because it's specified as an argument to the constructor [here](https://github.com/Financial-Times/dotcom-tool-kit/blob/main/lib/types/src/index.ts#L189)

ALTHOUGH, I realised during the demo this morning, and have been reminded now, that now that the `job` isn't referenced anywhere in the abstract class we can make it a `static` field which means other classes can reference a class's job without instantiating an instance. new commit incoming...

lib/types/src/circleci.ts
```diff
@@ -22,10 +23,42 @@ export type Workflow = {
 }
 
 export interface CircleConfig {
-  version: number
-  orbs: { [orb: string]: string }
-  workflows?: {
-    version: number
-    [workflow: string]: Workflow | number
+  version: 2.1
```
jkerr321, line 26: **suggestion:** I think this type is only used in the `circleci-config.ts` file, where [`PartialDeep` is applied to it ](https://github.com/Financial-Times/dotcom-tool-kit/pull/359/files#diff-f157d7392cc3db08e0c5b213317a296767e475221a5be30ae3bcc8b94fa7613fR25) which makes every field optional. Could you also add a comment in this file to note that actually every field is optional (and how/where that's done) so that it's not confusing to someone reviewing the types in isolation

plugins/circleci/src/circleci-config.ts
ivomurrell: I've been exporting a `*State` type for [each plugin](https://github.com/Financial-Times/dotcom-tool-kit/blob/69904edc5e01f484e836289932c5378ba9ec9f36/lib/package-json-hook/src/package-json-helper.ts#L18) that uses the shared install state, in case other packages want to reference the state type. `CircleCIState` was previously a [custom interface](https://github.com/Financial-Times/dotcom-tool-kit/blob/69904edc5e01f484e836289932c5378ba9ec9f36/plugins/circleci/src/circleci-config.ts#L14), but now happens to be an alias to another type. In the future we could add additional fields to `CircleCIState` type that are not included in the `CircleConfig` type without causing a breaking change.

plugins/circleci/src/circleci-config.ts
```diff
@@ -1,29 +1,178 @@
...
+/**
+ * CircleCIStatePartial makes every property in the config object optional,
+ * including nested properties. This is particularly useful as we merge each
+ * hook's config into the larger state, and so each hook only needs to define
+ * the parts of the config they want to add to.
+ */
```
jkerr321, line 24: **praise:** I love how well documented this code is, I think it's really helpful with something as complex as this Thank you! 🙌

plugins/circleci/src/circleci-config.ts
```diff
@@ -1,29 +1,178 @@
...
+}
+
+/**
+ * `generateConfigWithJob` generates a single job, structured so that it will
+ * merge nicely with the rest of the config. This will include the `requires`
+ * parameter, as well as the boilerplate properties from `jobBoilerplate`, but
+ * any other options will need to be passed to `additionalFields`, such as
+ * `filters.branches`.
+ */
+export const generateConfigWithJob = (options: JobGeneratorOptions): CircleCIStatePartial => {
```
jkerr321, line 69: **praise:** this is a great function! Tidy, readable, and a helpful level of abstraction! Love it! `:dancing-otter:` 😁

lib/types/src/circleci.ts
```diff
@@ -22,10 +23,42 @@ export type Workflow = {
...
+  }
+  executors: {
+    [executor: string]: {
+      docker: { image: string }[]
+    }
+  }
+  jobs: {
+    [job: string]: {
+      docker: { image: string }[]
+      steps: (string | { [command: string]: { path?: string } })[]
```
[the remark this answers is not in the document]
ivomurrell: because the `path` field isn't necessarily used in each step in a complete CircleCI config, whereas, for instance, each `job` in a complete CircleCI config _must_ have a `steps` object. _but_ a hook doesn't need to define a `steps` object in each job, because maybe another hook defines that part

plugins/circleci-npm/src/index.ts
```diff
@@ -1,15 +1,18 @@
...
-      branches: { ignore: '/.*/' }
+    addToNightly: false,
+    additionalFields: {
+      context: 'npm-publish-token',
+      filters: {
+        branches: { ignore: '/.*/' }
+      }
     }
-  }
-  runOnVersionTags = true
```
jkerr321, line 12: **praise:** nice getting rid of this! Not having to evaluate this in all the places feels cleaners and more simple 😊

lib/types/src/circleci.ts
ivomurrell: Seeing as it's aliased to `CircleCIState` it is used in quite a few places in `circleci-config.ts` where `PartialDeep` is not applied (`PartialDeep` is only applied when using `CircleCIStatePartial` instead). So the fields are not optional unless you're using the `CircleCIStatePartial` type instead of `CircleCIState` or `CircleConfig`

jkerr321: Ahh I seeeeee

lib/types/src/circleci.ts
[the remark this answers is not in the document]
jkerr321: 🤯 😆 Ok I think I get it! Thank you!

jkerr321 (approved):
👯 You have my approval!

PR #361, opened 2023-02-22
CPP-1053 Add ServerlessProvision task with OIDC token support

ivomurrell:
# Description

This adds a task to the serverless plugin to deploy review apps to AWS via Serverless. The task itself tries to mirror the `provision` target present in the `Makefile`'s of projects still using n-gage to deploy to AWS, like [next-anubis-config-api](https://github.com/Financial-Times/next-anubis-config-api/blob/39f48d9d100d5d3d8fceeec0459e19d77e1f1a9d/Makefile#L24). The main point of interest is the introduction of using [OIDC tokens](https://tech.in.ft.com/tech-topics/continuous-delivery/circleci-aws-oidc-authentication#create-circleci-config-using-the-aws-cli-orb-in-your-project) to [deploy to CircleCI](https://circleci.com/docs/openid-connect-tokens/). I've had to add some additional logic to the plugin loading code and the Tool Kit orb in order to support this, as it requires the use of commands specified in the `circleci/aws-cli` and `circleci/serverless-framework` orbs to be run before the other commands in the `heroku-provision` orb job (pending a rename...) are. This is written on top of the work done recently to revamp the `circleci` plugin, though in the end we didn't actually need to use the ability to define multiple jobs for a single hook that the revamp now allows as I moved the problematic commands inside the orb itself 😅

# Checklist:

### DIFF
--- a/lib/types/src/circleci.ts
+++ b/lib/types/src/circleci.ts
@@ -7,6 +7,7 @@ export type JobConfig = {
   requires?: string[]
   filters?: { branches?: { only?: string; ignore?: string }; tags?: { only?: string } }
   executor?: string
+  [parameter: string]: unknown
 }
 
 type TriggerConfig = {
--- a/lib/types/src/schema.ts
+++ b/lib/types/src/schema.ts
@@ -39,6 +39,7 @@ import { NodeSchema } from './schema/node'
 import { NodemonSchema } from './schema/nodemon'
 import { Pa11ySchema } from './schema/pa11y'
 import { PrettierSchema } from './schema/prettier'
+import { ServerlessSchema } from './schema/serverless'
 import { TypeScriptSchema } from './schema/typescript'
 import { UploadAssetsToS3Schema } from './schema/upload-assets-to-s3'
 import { VaultSchema } from './schema/vault'
@@ -59,6 +60,7 @@ export const Schemas = {
   '@dotcom-tool-kit/nodemon': NodemonSchema,
   '@dotcom-tool-kit/pa11y': Pa11ySchema,
   '@dotcom-tool-kit/prettier': PrettierSchema,
+  '@dotcom-tool-kit/serverless': ServerlessSchema,
   '@dotcom-tool-kit/typescript': TypeScriptSchema,
   '@dotcom-tool-kit/upload-assets-to-s3': UploadAssetsToS3Schema,
   '@dotcom-tool-kit/vault': VaultSchema,
--- a/orb/src/jobs/heroku-provision.yml
+++ b/orb/src/jobs/heroku-provision.yml
@@ -2,11 +2,26 @@ parameters:
   executor:
     default: default
     type: executor
+  aws-account-id:
+    default: ''
+    type: string
+  system-code:
+    default: ''
+    type: string
 
 executor: << parameters.executor >>
 
 steps:
   - attach-workspace
+  - when:
+      condition:
+        and:
+          - << parameters.aws-account-id >>
+          - << parameters.system-code >>
+      steps:
+        - serverless-assume-role:
+            aws-account-id: << parameters.aws-account-id >>
+            system-code: << parameters.system-code >>
   - run:
       name: Create and test Heroku review app
       command: npx dotcom-tool-kit deploy:review
--- a/plugins/circleci-deploy/src/index.ts
+++ b/plugins/circleci-deploy/src/index.ts
@@ -1,15 +1,34 @@
-import CircleCiConfigHook, { generateConfigWithJob } from '@dotcom-tool-kit/circleci/lib/circleci-config'
+import CircleCiConfigHook, {
+  CircleCIStatePartial,
+  generateConfigWithJob,
+  JobGeneratorOptions
+} from '@dotcom-tool-kit/circleci/lib/circleci-config'
 import { TestCI } from '@dotcom-tool-kit/circleci/lib/index'
 import { getOptions } from '@dotcom-tool-kit/options'
+import type { SetRequired } from 'type-fest'
 
 export class DeployReview extends CircleCiConfigHook {
   static job = 'tool-kit/heroku-provision'
-  config = generateConfigWithJob({
-    name: DeployReview.job,
-    addToNightly: true,
-    requires: ['tool-kit/setup', 'waiting-for-approval'],
-    additionalFields: { filters: { branches: { ignore: 'main' } } }
-  })
+  // needs to be a getter so that we can lazily wait for the global options
+  // object to be assigned before getting values from it
+  get config(): CircleCIStatePartial {
+    // Use SetRequired to work around TypeScript thinking additionalFields
+    // could be undefined even though we've set the field here
+    const jobOptions: SetRequired<JobGeneratorOptions, 'additionalFields'> = {
+      name: DeployReview.job,
+      addToNightly: true,
+      requires: ['tool-kit/setup', 'waiting-for-approval'],
+      additionalFields: {
+        filters: { branches: { ignore: 'main' } }
+      }
+    }
+    const serverlessOptions = getOptions('@dotcom-tool-kit/serverless')
+    if (serverlessOptions?.awsAccountId && serverlessOptions?.systemCode) {
+      jobOptions.additionalFields['aws-account-id'] = serverlessOptions.awsAccountId
+      jobOptions.additionalFields['system-code'] = serverlessOptions.systemCode
+    }
+    return generateConfigWithJob(jobOptions)
+  }
 }
 
 export class DeployStaging extends CircleCiConfigHook {
--- a/plugins/serverless/src/tasks/provision.ts
+++ b/plugins/serverless/src/tasks/provision.ts
@@ -1,0 +1,55 @@
+import { ToolKitError } from '@dotcom-tool-kit/error'
+import { hookFork, styles } from '@dotcom-tool-kit/logger'
+import { Task } from '@dotcom-tool-kit/types'
+import { ServerlessSchema } from '@dotcom-tool-kit/types/lib/schema/serverless'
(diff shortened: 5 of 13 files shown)


plugins/serverless/src/tasks/provision.ts
```diff
@@ -0,0 +1,60 @@
...
+  static description = 'Provisions a job on AWS'
+
+  static defaultOptions = {
+    useVault: true,
+    buildNumVariable: 'CIRCLE_BUILD_NUM'
+  }
+
+  async run(): Promise<void> {
+    const { useVault, configPath } = this.options
+    const buildNum = process.env[this.options.buildNumVariable!]
```
joelcarr, line 18: **question:** what is the `!` for?

plugins/backend-serverless-app/src/index.ts
```diff
@@ -0,0 +1,32 @@
+import CircleCiConfigHook, {
+  CircleCIStatePartial,
+  generateConfigWithJob
+} from '@dotcom-tool-kit/circleci/lib/circleci-config'
+import { ToolKitError } from '@dotcom-tool-kit/error'
+import { getOptions } from '@dotcom-tool-kit/options'
+
+export class ServerlessDeployReview extends CircleCiConfigHook {
```
joelcarr, line 8: **question:** Why is this hook defined in here rather than the `@dotcom-tool-kit/serverless` plugin?

plugins/serverless/src/tasks/provision.ts
jkerr321: I think it solves this TS error:

image

Without the `!` TS evaluates the type as `string | undefined`; adding the `!` asserts that the value is never null or undefined and so TS accepts it.

jkerr321: ...although the next line of code is 
```
if (buildNum === undefined) {
```
sooooo it _can_ be undefined even though we're asserting to TS that it won't be?

lib/types/src/circleci.ts
```diff
@@ -7,6 +7,7 @@ export type JobConfig = {
   requires?: string[]
   filters?: { branches?: { only?: string; ignore?: string }; tags?: { only?: string } }
   executor?: string
+  [parameter: string]: unknown
```
jkerr321, line 10: **question:** what is this for?

lib/types/src/schema.ts
```diff
@@ -71,6 +71,7 @@ import type { BabelOptions } from './schema/babel'
 import type { CircleCIOptions } from './schema/circleci'
 import type { CypressOptions } from './schema/cypress'
 import type { TypeScriptOptions } from './schema/typescript'
+import type { ServerlessOptions } from './schema/serverless'
```
jkerr321, line 74: **praise:** ah thanks! I missed this in my initial serverless work.

plugins/serverless/src/tasks/run.ts
```diff
@@ -9,7 +9,7 @@ import waitPort from 'wait-port'
 export default class ServerlessRun extends Task<typeof ServerlessSchema> {
   static description = 'Run serverless functions locally'
 
-  static defaultOptions: ServerlessOptions = {
```
jkerr321, line 12: **question:** Why are you removing the `ServerlessOptions` type here? Is it in preparation for the Zod PR where this will no longer be needed?

plugins/serverless/src/tasks/provision.ts
joelcarr: Ahh I see! Thank you 🙏 

It looks like that `process.env[this.options.buildNumVariable]` can be undefined but we are saying the option itself cannot. And it looks like we make sure it is not undefined by defining a `defaultOption` of it above but TS hasn't made that connection.

orb/src/jobs/heroku-provision.yml
```diff
@@ -2,11 +2,26 @@ parameters:
...
+    type: string
+  system-code:
+    default: ''
+    type: string
 
 executor: << parameters.executor >>
 
 steps:
   - attach-workspace
+  - when:
```
jkerr321, line 16: **praise:** great commit message description of why you chose this solution. The code is tidy and readable so it def feels to me like the best approach 🎉

plugins/serverless/src/tasks/provision.ts
ivomurrell: Yeah exactly, we're adding the default options [here](https://github.com/Financial-Times/dotcom-tool-kit/blob/18e4a271c43711a1160faed4662f41b4c5329880/lib/types/src/index.ts#L159) but TypeScript doesn't know this (we could probably solve that with an extra type parameter but it would get quite verbose). zod solves this issue 😄

plugins/backend-serverless-app/src/index.ts
jkerr321: I'm also interested in the answer to this. I was going to reply to Joel that I thought it was because we needed to import the Circle plugin in order to override the hook and that maybe made more sense at the app level, but actually it could happen in the Serverless plugin too

plugins/serverless/src/tasks/provision.ts
jkerr321: 👏 great work team! 😁

jkerr321 (approved):
I have a couple of non-blocking questions but apart from that it looks GREAT!

plugins/backend-serverless-app/src/index.ts
ivomurrell: No that's right, Jen! The Serverless plugin is generic and shouldn't be aware of CircleCI so that we could support other CI systems in the future. The `backend-serverless-app` is where the serverless and circleci-deploy plugins get pulled in together so it makes sense to put the CircleCI hook for Serverless code there.

lib/types/src/circleci.ts
ivomurrell: Any [parameters](https://circleci.com/docs/configuration-reference/#parameters-job) you've specified in a job definition can be passed to the job. These parameters can be any string (excluding some reserved words and only with valid characters) which we represent with an [index signature](https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures). For example, the Tool Kit `heroku-provision` job now has takes [two new parameters](https://github.com/Financial-Times/dotcom-tool-kit/blob/9f3c4be57255fd39b87aed801218cc00c7769c65/orb/src/jobs/heroku-provision.yml#L5-L10), which we use to [pass settings from your options](https://github.com/Financial-Times/dotcom-tool-kit/blob/9f3c4be57255fd39b87aed801218cc00c7769c65/plugins/backend-serverless-app/src/index.ts#L23-L24).

plugins/serverless/src/tasks/provision.ts
```diff
@@ -0,0 +1,60 @@
...
+import { hookFork, styles } from '@dotcom-tool-kit/logger'
+import { Task } from '@dotcom-tool-kit/types'
+import { ServerlessSchema } from '@dotcom-tool-kit/types/lib/schema/serverless'
+import { VaultEnvVars } from '@dotcom-tool-kit/vault'
+import { spawn } from 'child_process'
+
+export default class ServerlessProvision extends Task<typeof ServerlessSchema> {
+  static description = 'Provisions a job on AWS'
+
+  static defaultOptions = {
```
jkerr321, line 11: **issue:** just realised the [readme](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/serverless) needs to be updated with the new task and options

plugins/serverless/src/tasks/run.ts
ivomurrell: No, it's because it throws a type error if we keep it as we don't specify default options for [`awsAccountId`, `systemCode`, and `region`](https://github.com/Financial-Times/dotcom-tool-kit/blob/9f3c4be57255fd39b87aed801218cc00c7769c65/lib/types/src/schema/serverless.ts#L4-L6), which are required options but must be specified by the user. I was allowing it to fall back to the default type for `defaultOptions`, [`Record<string, unknown>`](https://github.com/Financial-Times/dotcom-tool-kit/blob/9f3c4be57255fd39b87aed801218cc00c7769c65/lib/types/src/index.ts#L151), but I think it would be better to use `Partial<ServerlessOptions>` actually so that we can type check the fields we _are_ setting!

plugins/backend-serverless-app/src/index.ts
ivomurrell: We could handle it so that we check for the `@dotcom-tool-kit/serverless` options in the `circleci-deploy` plugin but this feels a little weird to me. Though maybe it's okay? I'll try it out

jkerr321: 👍 I'm trying to think about it in terms of a new developer coming to Tool Kit in a year's time and it being as obvious and predictable as possible why things happen where they do. I don't know that my suggestion is the better solution but I appreciate you considering it and having a go!

plugins/serverless/src/tasks/run.ts
jkerr321: Nice!

plugins/circleci-deploy/src/index.ts
```diff
@@ -1,15 +1,34 @@
...
-  config = generateConfigWithJob({
-    name: DeployReview.job,
-    addToNightly: true,
-    requires: ['tool-kit/setup', 'waiting-for-approval'],
-    additionalFields: { filters: { branches: { ignore: 'main' } } }
-  })
+  // needs to be a getter so that we can lazily wait for the global options
+  // object to be assigned before getting values from it
+  get config(): CircleCIStatePartial {
+    // Use SetRequired to work around TypeScript thinking additionalFields
```
jkerr321, line 15: **praise:** thanks for adding a comment explaining why this is used

jkerr321 (approved):
Nice simplification, thank you!

PR #363, opened 2023-03-10
build!: revise Cypress base URL value for staging scenario

andygout:
# Description

This PR changes the value assigned to `cypressEnv.CYPRESS_BASE_URL` in the staging scenario so that the the environment variable contains the entire URL rather than just the hostname.

The rationale is that this will require a new key type added to [cp-vault-interface](https://github.com/Financial-Times/cp-vault-interface) for `heroku-url` (done in this [branch](https://github.com/Financial-Times/cp-vault-interface/compare/add-heroku-url-key-type)). The other `xxxxx-url` validators in cp-vault-interface (currently only `raven-url` and `slack-webhook-url`) use the [URL validator](https://github.com/Financial-Times/cp-vault-interface/blob/main/lib/validators/key-types/url.js).

From a consistency perspective it would make sense that all URL values in Vault are the full URL rather than just the hostname.

We can also validate a) that the value is a URL, b) that its protocol is `https:`, and c) its format matches the Heroku hostname format, as long as the value is updated using cp-vault-interface rather than manually, though admittedly this is dependent on individuals and therefore not a guarantee while the hardcoded `https:` in the Cypress plugin's existing state does enforce the protocol, so there is certainly an argument as to whether this is the best approach (both here and more generally).

I have added @Financial-Times/cp-reliability as reviewers to this PR as I think it feeds into this larger discussion of secret management.

#### Naming change for the environment variable
I changed the name from `CY_` to `CYPRESS_` so as to make explicit the purpose of the environment variable as those less familiar with Cypress may not be aware of its `cy` usage shorthand.

#### Semver release decision
I have used a `!` in the commit to indicate a breaking change. While I don't think there are any apps that have yet migrated to Tool Kit and are running end-to-end Cypress tests against the staging app in their CircleCI production deployment pipeline, I'm not certain this is definitely the case and so in the interests of caution it should be a major release as the `tool-kit/e2e-test-staging` job will fail at **Run smoke tests on staging** without the necessary environment variable amends.

# Checklist:


andygout:
**Update:** The cp-vault-interface [PR](https://github.com/Financial-Times/cp-vault-interface/pull/113) was closed for reasons described by the discussion (non-secret values are not subject to validation), meaning this PR comes down to the points of:
- consistency of having full URLs in Vault (i.e. including the protocol)
- explicitness of the env var name

PR #365, opened 2023-03-14
Validate options even when none are given for a plugin 

ivomurrell:
# Description

Now that we've moved the setting of default options into the zod schema validation, we need to make sure we're using zod to parse options for plugins even when no options have been specified explicitly in your `.toolkitrc.yml` config. This is also important to ensure that plugins with required options are having those set.

For example, @andygout was encountering an [issue](https://financialtimes.slack.com/archives/C3TJ6KXEU/p1678702462886019) with the latest Tool Kit versions where the `run:local` hook was failing to run with an error
```
Nodemon:
Cannot destructure property ‘entry’ of ‘this.options’ as it is undefined.
```
This is because `nodemon` has a default value [specified](https://github.com/Financial-Times/dotcom-tool-kit/blob/5a450eabee673c8a4ceecf72f6eba7bf2e9dfcd2/lib/types/src/schema/nodemon.ts#L4) for the `entry` option, but seeing as they [weren't explicitly setting](https://github.com/Financial-Times/next-newsletter-signup/blob/e9a8cc11e8c531fdcbe32239a3372cc470fc1368/.toolkitrc.yml#L31) any options for `@dotcom-tool-kit/nodemon`, the plugin's option schema would never be loaded, and therefore zod would never have a chance to parse the (empty) option object and return a validated object with default values set for unspecified fields.

I've now updated the configuration validation code to iterate over the loaded plugins rather than the loaded options. This means we won't miss any plugins that haven't had any options explicitly set. Note that we won't miss any options that aren't associated with a plugin, as unused options were already disallowed. Any options that are created implicitly like this will be linked to the user's root configuration rather than a nested `.toolkitrc.yml` file.

# Checklist:

### DIFF
--- a/core/cli/src/config.ts
+++ b/core/cli/src/config.ts
@@ -137,23 +137,35 @@ export function validateConfig(config: ValidPluginsConfig, logger: Logger): Vali
   }
 
   const invalidOptions: InvalidOption[] = []
-  for (const [id, options] of Object.entries(config.options).filter((entry): entry is [
-    string,
-    PluginOptions
-  ] => {
-    const [, options] = entry
-    return !!options && !isConflict(options)
-  })) {
+  for (const [id, plugin] of Object.entries(config.plugins)) {
     const pluginId = id as keyof SchemaOptions
+    const pluginOptions = config.options[pluginId]
+    if (pluginOptions && isConflict(pluginOptions)) {
+      continue
+    }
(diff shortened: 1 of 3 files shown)


joelcarr (approved):
⚡

joelcarr (approved):

PR #366, opened 2023-03-14
CPP-1353: add upload-assets-to-s3 plugin to frontend app

joelcarr:
We include webpack in the frontend-app bootstrap plugin to compile frontend assets. We should include upload-assets-to-s3 so those assets are deloyed in production to the asset s3 bucket.

# Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,13 +45,13 @@
     },
     "core/cli": {
       "name": "dotcom-tool-kit",
-      "version": "2.6.0",
+      "version": "2.6.1",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


apaleslimghost (approved):

andygout:
Ah, nice - will this mean I can remove declaring the `upload-assets-to-s3` plugin explicitly in the `.toolkitrc.yml`? https://github.com/Financial-Times/next-newsletter-signup/pull/378/files

joelcarr:
@andygout that is right. It means that it will upload assets for all frontend apps just like n-gage used to.

PR #369, opened 2023-03-15
CPP-1293: add husky as a dependency to @dotcom-tool-kit/husky-npm plugin

joelcarr:
When a repo installs to @dotcom-tool-kit/husky-npm we expect husky to be bundled so that our pre-commit hook can execute. However it turns out that when husky is defined as a peer dependency, although it does get installed, its executable is not defined in the root node_modules/.bin folder. This means that the `husky-run` executable cannot be found by the git pre-commit hook.

The plugin should depend on husky. This matches the way it worked for n-gage.

# Checklist:

### DIFF
--- a/plugins/husky-npm/package.json
+++ b/plugins/husky-npm/package.json
@@ -11,10 +11,10 @@
   "license": "ISC",
   "dependencies": {
     "@dotcom-tool-kit/package-json-hook": "^3.0.0",
-    "tslib": "^2.3.1"
+    "tslib": "^2.3.1",
+    "husky": "^4.3.8"
   },
   "peerDependencies": {
-    "husky": "4.x",
     "dotcom-tool-kit": "^2.6.0"
   },
   "repository": {


ivomurrell (commented):
**issue**: how do we make sure that a user has our version of husky hoisted to the root of `node_modules` and not another version that isn't compatible with our `husky` plugin? other versions of husky define their configuration in different places. we depend on the fact that `husky@4` has a `prepare` script to automatically install itself into your git hooks, but now that we don't have it specified as a peer dependency there's no make sure that something else isn't pulling in a competing major version of husky.

joelcarr:
While this may be an issue, we didn't defend against it for [@financial-times/n-gage](https://github.com/Financial-Times/n-gage/blob/1edbe16b1f49058b534c0bd117ce5bd0f17d13c2/package.json#L31). If a repo installs a newer version of `husky`, they will have moved away from our supported version and the maintainers of the repo will have to accept they have to manage the config themselves.

We should consider supporting the latest version of husky. It looks like it sets a git config `hooksPath` to the `.husky` directory which is installed to the root of the repo which doesn't look too complex to implement. What do you think?

PR #372, opened 2023-03-17
CPP-1054: add ServerlessDeploy task to @dotcom-tool-kit/serverless

joelcarr:
# Description

Add a `ServerlessDeploy` task to `@dotcom-tool-kit/serverless`. This also changes the `region` property type to be a array of regions as there are platinum services that need to be deployed to EU West and US East.

# Checklist:


lib/types/src/schema/serverless.ts
```diff
@@ -3,11 +3,11 @@ import { z } from 'zod'
 export const ServerlessSchema = z.object({
   awsAccountId: z.string(),
   systemCode: z.string(),
-  region: z.string(),
+  region: z.array(z.string()).default(['eu-west-1']),
```
ivomurrell, line 6: **issue**: seeing as the option schema is changing this needs to be a breaking change denoted by a commit message with an exclamation mark after `feat(serverless)`. this isn't too bad seeing as we're doing prerelease versions atm so it (should) only jump to `0.2.0`.

plugins/serverless/src/tasks/deploy.ts
```diff
@@ -0,0 +1,49 @@
...
+import { Task } from '@dotcom-tool-kit/types'
+import { ServerlessSchema } from '@dotcom-tool-kit/types/lib/schema/serverless'
+import { VaultEnvVars } from '@dotcom-tool-kit/vault'
+import { spawn } from 'child_process'
+
+export default class ServerlessDeploy extends Task<typeof ServerlessSchema> {
+  static description = 'Deploys on AWS'
+
+  async run(): Promise<void> {
+    const { useVault, configPath } = this.options
```
ivomurrell, line 12: **nitpick**: which variables are deconstructed from `this.options` are a little inconsistent here, we can probably extract the `buildNumVariable` and `region` variables too. it was an oversight in the `provision.ts` file too as I added the `buildNumVariable` option later on 😅

ivomurrell (commented):
a couple of small issues here but this looks good, thank you for taking the time to do some Serverless work!!

ivomurrell (commented):
**thought**: this is a breaking change to the `types` library which seems a bit overkill seeing as it only affects this plugin but it is correct. perhaps we should think about situating the option schemas in the plugins' packages instead of shared in `types` in the future to avoid this?

ivomurrell (approved):

PR #379, opened 2023-03-27
Force color output for mocha logs

ivomurrell:
# Description

As with many CLI tools, mocha supports ANSI colours for its test reporting but disables them if it's not connected to a TTY (i.e., if it's not an interactive session, such as when it's run in a CI or if it's being piped to another script). It doesn't detect a TTY in our case as we're piping its stdout and stderr through our winston logging infrastructure rather than outputting to the user's terminal directly. In order to get colourful output then, let's force it via mocha's [`--color`](https://mochajs.org/#-color-c-colors) option.

# Checklist:


jkerr321 (approved):
Noice

PR #383, opened 2023-03-28
CPP-1046 Clarify job names in orb

ivomurrell:
# Description

The deployment jobs defined in the Tool Kit orb previously had names specifically tied to their initially conceived use case: `heroku-promote` and `heroku-staging`, for example. These jobs should be equally suitable to use for many different kinds of deployment – such as deployments via Serverless – and, in fact, the Tool Kit hooks they call are already designed to much more generic (`deploy:production` and `deploy:staging` respectively). This can be confusing for people extending Tool Kit as they may avoid these jobs when writing CircleCI hooks when in reality they are perfectly appropriate.

We are also renaming the `publish` job which suffers from the opposite problem – its name is too generic. It’s actually a job to publish changes to npm and is separated from the other jobs as it uses a different hook, `publish:tag`.

# Checklist:


orb/src/jobs/heroku-promote.yml
joelcarr, line 1: **question:** Do we have a plan to deprecate and remove these old heroku specific jobs?

ivomurrell: we can remove them in a future major release of the orb. perhaps before that we could add a new step in the jobs to print a deprecation warning?

joelcarr: Made this placeholder ticket https://financialtimes.atlassian.net/browse/CPP-1369

joelcarr (approved):

PR #412, opened 2023-04-19
CPP-1248 Use standard tsconfig base for Node 16

ivomurrell:
# Description

This sets the `target` and `library` compiler options so that we aren't polyfilling any ECMAScript constructs that Node 16 supports natively, and the latest Node 16 library functions are type checked properly. It will be easier to remember to update this base configuration once we move to Node 18 than updating the options directly. These base options are defined [here](https://github.com/tsconfig/bases/blob/821d9aa0b2e075949b6764c1e4556f83ab7554a9/bases/node16.json).

It also sets some useful recommended default options, including some that we were already using which I've now removed as they're redundant.

# Checklist:

### DIFF
--- a/tsconfig.settings.json
+++ b/tsconfig.settings.json
@@ -1,14 +1,9 @@
 {
+  "extends": "@tsconfig/node16/tsconfig.json",
   "compilerOptions": {
     "composite": true,
     "declaration": true,
     "declarationMap": true,
-    "importHelpers": true,
-    "module": "commonjs",
-    "strict": true,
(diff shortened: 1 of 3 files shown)


tsconfig.settings.json
```diff
@@ -1,14 +1,9 @@
 {
+  "extends": "@tsconfig/node16/tsconfig.json",
```
joelcarr, line 2: **praise:** This is SO much nicer!

joelcarr (approved):

PR #415, opened 2023-04-25
CPP-1285 Improve Heroku errors

ivomurrell:
# Description

We've long had a problem where errors thrown when getting a unsuccessful status code from the Heroku API neither include the actual error response body from Heroku, nor provide any information on what API call was the one that caused the error. The `heroku-client` package itself is largely to blame for this: it does bundle up a good amount of information in the errors it throws, but it's hidden away in added fields to the `Error` class that the client doesn't document anywhere.

For this PR, I've augmented the errors thrown from the Heroku API to include the response body from the API, as well as requiring code to pass an argument describing what it's going to use the Heroku client for before you can use the client managed by `herokuClient.ts`. This means we can always provide additional context for where the Tool Kit task ran into issues. In the future it should be trivial to also add another (optional) argument that allows callers to pass a string on suggested remediations based on guesses on why we might have hit an error at a given call site, but I've spent enough time on this PR (particularly wrangling Jest's confusing mocking system) and don't particularly have the context necessary to write up these suggestions 😅

**Before:**
Screenshot 2023-04-25 at 12 20 34

**After:**
Screenshot 2023-04-25 at 12 19 48

# Checklist:

### DIFF
--- a/plugins/heroku/src/buildHerokuReviewApp.ts
+++ b/plugins/heroku/src/buildHerokuReviewApp.ts
@@ -1,25 +1,23 @@
-import heroku from './herokuClient'
+import heroku, { extractHerokuError } from './herokuClient'
 import type { HerokuApiResGetReview, HerokuApiResPost } from 'heroku-client'
 import { Logger } from 'winston'
 import { ToolKitError } from '@dotcom-tool-kit/error'
 import { repeatedCheckForSuccessStatus } from './repeatedCheckForSuccessStatus'
 import { getRepoDetails } from './githubApi'
 
-async function buildHerokuReviewApp(
-	logger: Logger,
-	pipelineId: string
-	): Promise<string> {
-
+async function buildHerokuReviewApp(logger: Logger, pipelineId: string): Promise<string> {
   const { branch, repo, source_blob } = await getRepoDetails(logger)
 
   logger.info(`creating review app for ${branch} branch on ${repo}...`)
-  const reviewAppBuild: HerokuApiResPost = await heroku.post(`/review-apps`, {
-    body: {
-      branch: branch,
-      pipeline: pipelineId,
-      source_blob
-    }
-  })
+  const reviewAppBuild = await heroku
+    .post<HerokuApiResPost>(`/review-apps`, {
+      body: {
+        branch: branch,
+        pipeline: pipelineId,
+        source_blob
+      }
+    })
+    .catch(extractHerokuError(`creating review app on pipeline ${pipelineId}`))
 
   logger.info('reviewApp in buildHerokuReviewApp', reviewAppBuild)
 
@@ -33,8 +31,10 @@ async function buildHerokuReviewApp(
                     The review-app build request was attempted on repo: ${repo}, branch: ${branch}, version: ${source_blob.version}.`
     throw error
   }
-  
-  const reviewApp: HerokuApiResGetReview = await heroku.get(`/review-apps/${reviewAppBuild.id}`)
+
+  const reviewApp: HerokuApiResGetReview = await heroku
+    .get<HerokuApiResGetReview>(`/review-apps/${reviewAppBuild.id}`)
+    .catch(extractHerokuError(`getting app ID for review app ${reviewAppBuild.id}`))
 
   return reviewApp.app.id
 }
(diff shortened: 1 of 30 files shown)


plugins/heroku/src/herokuClient.ts
```diff
@@ -1,5 +1,30 @@
...
+): Promise<T> {
+  try {
+    return await callback(client)
+  } catch (e) {
+    const upstream = e as HerokuError
+    const error = new ToolKitError('there was an error calling the Heroku API')
+    error.details = `received the error ${styles.errorHighlight(upstream.message)} when ${context}`
+    if (upstream.body) {
+      error.details += `\n\nresponse from Heroku was:\n${highlight(
+        JSON.stringify(upstream.body, undefined, 2),
```
apaleslimghost, line 24: **suggestion**: Heroku errors always have `id` and `message` fields, and sometimes a `url`, can we just print the `message` and `url`? think it would be cleaner than outputting JSON. the `id` seems to be just [based on the status code](https://devcenter.heroku.com/articles/platform-api-reference#client-error-responses), so we can skip that.

plugins/heroku/src/herokuClient.ts
```diff
@@ -1,5 +1,30 @@
...
 
 const HEROKU_AUTH_TOKEN = process.env.HEROKU_AUTH_TOKEN
 
-export default new Heroku({ token: HEROKU_AUTH_TOKEN })
+const client = new Heroku({ token: HEROKU_AUTH_TOKEN })
+
+// the Heroku client doesn't give particularly useful error messages, so we require an additional context
+// string to be included when calling the API which we include in custom error handling
+export async function herokuWithContext<T>(
+  context: string,
```
apaleslimghost, line 13: **suggestion**: instead of passing in a string context with this call, could we extract the error handling into a separate function, to be used like <code>.catch(handleHerokuError(\`promoting app ID ${appId} to production`))</code>? would be cleaner than the callback wrapper

apaleslimghost (approved):
looks good, a few nonblocking suggestions. it would be nice to see more context about why the errors could have happened, maybe we should put some time in to go through that as a group?

plugins/heroku/src/herokuClient.ts
ivomurrell: the reason I went with this approach was to _force_ a context be applied when using the `heroku-client`. your alternative is definitely cleaner but I worry people will forget to use it if we add new Heroku API calls to the plugin. what do you think?

apaleslimghost: that's a good point, but i think having the callback makes it a bit non-obvious what this function is actually doing when you're reading a call? i don't forsee us changing the `heroku` plugin much, and imo we should be more diligent about things like error messages in reviews anyway

plugins/heroku/src/buildHerokuReviewApp.ts
```diff
@@ -1,25 +1,23 @@
...
   logger.info(`creating review app for ${branch} branch on ${repo}...`)
-  const reviewAppBuild: HerokuApiResPost = await heroku.post(`/review-apps`, {
-    body: {
-      branch: branch,
-      pipeline: pipelineId,
-      source_blob
-    }
-  })
+  const reviewAppBuild = await heroku
+    .post<HerokuApiResPost>(`/review-apps`, {
```
apaleslimghost, line 13: **praise**: nice work adding these generics to all the calls, that'll definitely help maintainability of this plugin

plugins/heroku/src/herokuClient.ts
```diff
@@ -1,5 +1,27 @@
...
 
 const HEROKU_AUTH_TOKEN = process.env.HEROKU_AUTH_TOKEN
 
 export default new Heroku({ token: HEROKU_AUTH_TOKEN })
+
+// the Heroku client doesn't give particularly useful error messages, so we require an additional context
+// string when handling errors coming from it
+export function extractHerokuError(context: string): (err: unknown) => never {
+  return (err) => {
+    if ((err as HerokuError).statusCode) {
```
apaleslimghost, line 13: **suggestion**: this would be safer as an `instanceof HerokuError` check (which would also save you from having to do `const upstream = err as HerokuError`)

apaleslimghost (approved):
love it, one minor suggestion

plugins/heroku/src/buildHerokuReviewApp.ts
ivomurrell: no need to praise me, tsc fails if you don't put them here! 😅 previously it could infer the type parameter based on the type of the variable we're assigning the return value to, but now that we weave the result through a `catch` statement too the type inference fails, even though `extractHerokuError` is diverging and therefore doesn't change the return value

PR #416, opened 2023-04-25
CPP-1335 CPP-1336 Support running Tool Kit with Node 18 (NOT supporting using Node 18 in Tool Kit projects)

ivomurrell:
# Description

This PR does the necessary groundwork to start running and testing Tool Kit with Node 18. The pinned Volta version is Node 18 (and the `npm@9` version that is bundled with it), updating the `engines` field to officially support it, and running CI builds against it. **This does not mean that we now support updating projects that use Tool Kit to Node 18**. Instead, this is just the first step in ensuring we can support that by making sure the code is compatible with the new runtime version.   

# Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ references:
       - image: cimg/node:<< parameters.node-version >>
     parameters:
       node-version:
-        default: '16.14'
+        default: '18.16'
         type: string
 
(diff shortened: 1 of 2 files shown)


joelcarr:
That PR title 😆

jkerr321 (approved):
Niiice niiice

PR #417, opened 2023-04-26
CPP-1337 Disable native fetch in forked Node processes

ivomurrell:
# Description

We aren't ready to support native `fetch` yet as we're still using `node-fetch`/`isomorphic-fetch` in a lot of places and there will likely be subtle bugs we'll have to hunt for as part of a concentrated migration. In the meantime, to facilitate the migration to Node 18, Tool Kit will run its forked Node processes (i.e., `npm start`'ing Customer Products apps) with the `--no-experimental-fetch` flag set so that we can make the most of Node 18 whilst excluding the native `fetch` implementation. Passing this flag to a Node runtime that doesn't know about native `fetch` will result in the error
```
bad option: --no-experimental-fetch
```
so we need to use the builtin [`process.allowedNodeEnvironmentFlags`](https://nodejs.org/dist/latest-v18.x/docs/api/process.html#processallowednodeenvironmentflags) function to check that the flag is compatible first. If it isn't then the runtime doesn't support native `fetch` anyway (for now: it's possible a future version of Node will force native `fetch`, but we can say that this version of Tool Kit doesn't support that hypothetical version.)

**question**: Do we want to allow users to disable this flag if they've already migrated to native `fetch`? This might help with our own migrations...?

# Checklist:

### DIFF
--- a/plugins/node/src/tasks/node.ts
+++ b/plugins/node/src/tasks/node.ts
@@ -12,7 +12,7 @@ export default class Node extends Task<typeof NodeSchema> {
   static description = ''
 
   async run(): Promise<void> {
-    const { entry, args, useVault, ports } = this.options
+    const { entry, args, useVault, ports, allowNativeFetch } = this.options
 
     let vaultEnv = {}
 
@@ -38,6 +38,12 @@ export default class Node extends Task<typeof NodeSchema> {
       throw error
     }
 
+    let { execArgv } = process
+    // disable native fetch if supported by runtime
+    if (!allowNativeFetch && process.allowedNodeEnvironmentFlags.has('--no-experimental-fetch')) {
+      execArgv = [...execArgv, '--no-experimental-fetch']
+    }
+
     this.logger.verbose('starting the child node process...')
     const child = fork(entry, args, {
       env: {
@@ -45,6 +51,7 @@ export default class Node extends Task<typeof NodeSchema> {
         PORT: port.toString(),
         ...process.env
       },
+      execArgv,
       silent: true
     })
     hookFork(this.logger, entry, child)
--- a/plugins/nodemon/src/tasks/nodemon.ts
+++ b/plugins/nodemon/src/tasks/nodemon.ts
@@ -11,7 +11,7 @@ export default class Nodemon extends Task<typeof NodemonSchema> {
   static description = ''
 
   async run(): Promise<void> {
-    const { entry, configPath, useVault, ports } = this.options
+    const { entry, configPath, useVault, ports, allowNativeFetch } = this.options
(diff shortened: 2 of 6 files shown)


plugins/node/src/tasks/node.ts
```diff
@@ -38,13 +38,20 @@ export default class Node extends Task<typeof NodeSchema> {
       throw error
     }
 
+    const { execArgv } = process
+    // disable native fetch if supported by runtime
+    if (process.allowedNodeEnvironmentFlags.has('--no-experimental-fetch')) {
```
apaleslimghost, line 43: **praise**: nice, i didn't know this existed, thought we'd have to resort to version sniffing

plugins/nodemon/src/tasks/nodemon.ts
```diff
@@ -36,7 +36,12 @@ export default class Nodemon extends Task<typeof NodemonSchema> {
       PORT: port.toString(),
       ...process.env
     }
-    nodemon({ script: entry, env, stdout: false, configFile: configPath })
+    const config: nodemon.Settings = { script: entry, env, stdout: false, configFile: configPath }
+    // disable native fetch if supported by runtime
+    if (process.allowedNodeEnvironmentFlags.has('--no-experimental-fetch')) {
+      config.execMap = { js: 'node --no-experimental-fetch' }
```
apaleslimghost, line 42: **suggestion**: in `node` we're forwarding the parent's `execArgv`, do we want to do that here as well?

apaleslimghost (approved):
looks good. i would prefer allowing an escape hatch option to not set this flag, i wouldn't want consumers to be stuck having to copypasta an entire plugin if they wanted to use native fetch

plugins/nodemon/src/tasks/nodemon.ts
ivomurrell: it's not the default behaviour of `nodemon` to forward the `execArgv` so idk I think that might be surprising?

plugins/node/src/tasks/node.ts
```diff
@@ -38,13 +38,20 @@ export default class Node extends Task<typeof NodeSchema> {
...
+      execArgv.push('--no-experimental-fetch')
```
apaleslimghost, line 44: **question**: there might be side effects of mutating `process.execArgv`, should we clone the array?

ivomurrell: aaah yes good point!! lovely javascript

apaleslimghost (approved):

jkerr321 (approved):
🎉 fabulous!

PR #418, opened 2023-04-27
Make npm regex for determining release tags stricter

ivomurrell:
# Description

The terminating anchor was seemingly removed accidentally in a [previous commit](https://github.com/Financial-Times/dotcom-tool-kit/commit/72d2faa644b8c0e1c71131e3559e802017f0d0f5), but it's necessary to make sure prerelease versions don't use the `latest` tag on npm. e.g., `1.0.0-beta.1` would match the previous regex as the `1.0.0` substring would match, however it wouldn't match the new regex as the string continues afterwards, so the `$` does not match.

# Checklist:


apaleslimghost (approved):
my b

PR #421, opened 2023-04-28
CPP-1340 Begin supporting Node 18 for Tool Kit

ivomurrell:
# Description

I didn't realise but we only have an `engines` field in the root `package.json`, not for each package. This means that the `engines` field wasn't published for individual packages on npm. As the final part of officially adding Node 18 support for Tool Kit lets add this field so that it's explicit that we support Node 18. This also doubles as a way to get a minor version bump for each package, implicitly delineating when support was added.

# Checklist:


joelcarr (commented):
**question(non-blocking):** should we update https://github.com/Financial-Times/dotcom-tool-kit/blob/7df2a3a07e73a3f48121861d13001dd21632ec34/scripts/create-plugin.js#L26-L39 as well? Not sure if the script gets used 😂

jkerr321 (approved):
🎉 🥳

joelcarr (approved):
🍾

PR #428, opened 2023-05-05
feat: use latest version of n-test

jkerr321:
there is a beta version of n-test which supports node.js 18 so let's use that in the node.js 18 supporting version of the n-test plugin

# Checklist:

### DIFF
--- a/plugins/n-test/package.json
+++ b/plugins/n-test/package.json
@@ -13,7 +13,7 @@
     "@dotcom-tool-kit/logger": "^3.1.0",
     "@dotcom-tool-kit/state": "^3.1.0",
     "@dotcom-tool-kit/types": "^3.1.0",
-    "@financial-times/n-test": "^5.0.1",
+    "@financial-times/n-test": "^6.1.0-beta.1",
     "tslib": "^2.3.1"
   },
   "repository": {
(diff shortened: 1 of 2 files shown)


plugins/n-test/package.json
```diff
@@ -13,7 +13,7 @@
     "@dotcom-tool-kit/logger": "^3.1.0",
     "@dotcom-tool-kit/state": "^3.1.0",
     "@dotcom-tool-kit/types": "^3.1.0",
-    "@financial-times/n-test": "^5.0.1",
+    "@financial-times/n-test": "6.1.0-beta.1",
```
ivomurrell, line 16: **suggestion**: prepending the range with a caret will allow us to automatically pull in later versions of `6.x` once it's out of beta (e.g., `6.1.0`, `6.1.3`, `6.4.0`, etc) we'll still have to do it via `npm update` to bump the lockfile but the intention of the version range will be clearer 🙂
```suggestion
    "@financial-times/n-test": "^6.1.0-beta.1",
```

ivomurrell: It would be weird to release a version in parallel tbh – `6.1.0-beta.1` could (normally) alternatively be read as 'the first beta for the `6.1.0`' so I don't thiiiink you should do a release of `6.1.0` in parallel? But pinning the beta version (i.e., no caret) does make sense if you think that's a possibility!

jkerr321: Ahh ok, thanks! That makes sense

PR #430, opened 2023-05-05
CPP-1439 improve errors for CircleCI installation failures

ivomurrell:
# Description

Previously, we’d enforce that your CircleCI configuration was installed properly by checking if all the Tool Kit hooks were mentioned. This was quite unreliable as we’d sometimes want other parts of the configuration updated (such as the orb version, which we ended up carving out a special case for) but the update was never triggered because the hooks were still present and accounted for. This was changed as part of the rearchitecting of the plugin, and we now do a much more thorough check, ensuring that the configuration is a superset of the expected configuration we calculate in the plugin. However, the error handling was largely left as is, meaning it assumes that if the installation check has failed then its because you were missing Tool Kit hooks.

Now, we use the same diffing code that Jest uses for failed assertions to pretty-print what fields we were expecting were missing in the user's config. This fixes the issue with the previous approach where the error wouldn't print anything if the config was missing fields other than Tool Kit hooks. It also makes it much easier to see what changes need to be made to satisfy Tool Kit.

_before_:
Screenshot 2023-05-05 at 17 44 22
_after_:
Screenshot 2023-05-05 at 17 44 40
(etc.)

Note that we present the diff as if your config is a JSON object rather than a YAML file. This is to avoid issues with diffing raw YAML files and getting spurious diffs when fields are sorted differently.

# Checklist:


plugins/circleci/src/circleci-config.ts
```diff
@@ -266,29 +250,27 @@ If you would like a Tool Kit configured CircleCI config file to be generated for
...
+            delete received[rk]
+            continue
+          }
+          if (isPlainObject(rv)) {
+            removeExtraFields(expected[rk], rv)
+          }
+        }
+      }
+      removeExtraFields(state, config)
+      const difference = diff(state, config, { expand: false })
```
apaleslimghost, line 271: suggestion: could this YAML.stringify both objects and do a string diff? would avoid ordering issues that way

apaleslimghost (approved):

plugins/circleci/src/circleci-config.ts
ivomurrell: this is a good idea! but testing it now the diffs can get really confusing when arrays are out of order. the JS object diff is a bit messy too but a bit more parseable imo and I think is a bit better at creating a minimal diff.

uuuh here
Screenshot 2023-05-05 at 18 12 27
vs
Screenshot 2023-05-05 at 18 12 49

idk

magsallen (commented):
I'm not very hot on my tool kit, but this looks reasonable to me, is there anything blocking us from merging?

ivomurrell:
> is there anything blocking us from merging?

not particularly, there was just an open question of whether we present the diff as a JS object or a YAML file: https://github.com/Financial-Times/dotcom-tool-kit/pull/430#discussion_r1186327309 do you have a preference? 😁

I can merge this as soon as you've made a call (or said you don't mind) – I don't want to block your work further.

magsallen:
I've got a very slight preference for yaml as it's not as noisy but either would work just fine. Flip a coin?

ivomurrell:
Let's do YAML then! Will commit that, rebase and merge now

PR #431, opened 2023-05-05
Check that the base CircleCI configuration matches the user's too

ivomurrell:
# Description

We were mistakenly only checking that the config returned by each hook was a subset of the user's CircleCI config, but we additionally want to check that the base configuration included regardless of hooks also matches. This is so that we check that fields like the Tool Kit orb version or the node executor are correct.

# Checklist:


plugins/circleci/src/circleci-config.ts
```diff
@@ -196,8 +198,10 @@ export default abstract class CircleCiConfigHook extends Hook<CircleCIState> {
   _versionTag?: string
   abstract config: CircleCIStatePartial
 
-  async getCircleConfig(): Promise<string | undefined> {
+  async getCircleConfig(): Promise<[string | undefined, boolean]> {
+    let firstLoad = false
```
joelcarr, line 202: **nitpick:** This to me reads as _this is not the first load_ whereas we want to say _the first load hasn't happened yet_. To me I would name this `hasCheckedBaseConfig` to be super verbose matching the intent of this Boolean.

plugins/circleci/test/circleci-config.test.ts
```diff
@@ -18,10 +18,19 @@ jest.mock('fs', () => {
...
 
 describe('CircleCI config hook', () => {
-  class TestHook extends CircleCiConfigHook {
+  abstract class FakeCircleCiConfigHook extends CircleCiConfigHook {
+    async getCircleConfig(): Promise<[string | undefined, boolean]> {
+      const [config] = await super.getCircleConfig()
+      // pretend it's never the first time we've opened the config so that we
+      // skip the comparison with the base config generated by
+      // getInitialState()
+      return [config, false]
```
joelcarr, line 27: **chore:** We should add a test for when it has been the config has been checked once before.

plugins/circleci/src/circleci-config.ts
ivomurrell: I'm confused, it _does_ mean "this is not the first load", which is how you read it. Are you saying you want it to mean "the first load hasn't happened yet"?

joelcarr: Yes 👍 To me - as soon as `getCircleConfig()` is called, that is the first load so `firstLoad` should be true. I was not able to figure out the purpose/intent of `firstLoad` without parsing the code in my head. My aim for the nitpick is to make the variable name clearer so I can now its' intent without having to read the whole function and how it is used.

ivomurrell: I'd argue that we shouldn't change the variable to something like `hasCheckedBaseConfig` as `getCircleConfig` shouldn't concern itself with what callees are doing with the `firstLoad` (or whatever) boolean – it's a leaky abstraction, and might no longer be correctly named if we later add another piece of logic that uses the boolean.

joelcarr: Is there anyway we can make this clearer? I struggled to understand its' intent when first reading the block of code. The aim here is to make this easier to maintain.

ivomurrell: `isFirstLoad`? `isFirstLoadOfConfig`? I'm still struggling to understand what's unclear about it, sorry!

> as soon as getCircleConfig() is called, that is the first load so firstLoad should be true

this isn't true the majority of times it's called, only the first time it's called 😅

PR #432, opened 2023-05-15
CPP-1320 Support Node version matrices in CircleCI plugin

ivomurrell:
# Description

This PR adds the ability to run CI code with multiple versions of Node using CircleCI's [matrix jobs](https://circleci.com/docs/using-matrix-jobs/). This is useful when your project supports multiple different Node versions and you want to test that your code is compatible with all of them. To use the matrices you specify an array of Node versions for `@dotcom-tool-kit/circleci`'s `nodeVersion` option and the plugin will install a CircleCI executor for each and split each job that supports running multiple times (i.e., test and build jobs, but not deploy jobs). In order to maintain simplicity in the plugin (which benefits maintainers of both the Tool Kit repository and custom plugin authors), any job that supports matrices will always be passed a list of Node versions to run, even if the length of the list is just one. This should have no affect on the behaviour of the pipeline to how it worked previously, and means there are fewer conditional paths in the config generation logic.

I've tested that this code works in both a repository that uses a single Node version (next-static) and one that uses multiple (n-express). In both cases valid code was generated, as confirmed by the CircleCI CLI (running `circleci config validate .circleci/config.yml`).

Questions for the reviewer:
- The current implementation does not support the pipeline forking into matrix jobs, joining into a single job, then forking again. Will this ever be an issue?
- I figured that the `tool-kit/setup` job should be run in parallel, as `npm install` behaviour can be different based on the Node version – even when the npm version is the same (scripts that check `engines` for example). However, `npm install` is quite a long running process and it will be potentially expensive to run it multiple times. Do we think that this is acceptable or should we fall down on the tradeoff in the other direction?

# Checklist:

### DIFF
--- a/plugins/circleci/src/circleci-config.ts
+++ b/plugins/circleci/src/circleci-config.ts
@@ -15,7 +15,6 @@ import type { PartialDeep } from 'type-fest'
 import YAML from 'yaml'
 
 const MAJOR_ORB_VERSION = '4'
-const DEFAULT_NODE_VERSION = '16.14-browsers'
 
 export type CircleCIState = CircleConfig
 /**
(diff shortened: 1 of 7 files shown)


plugins/circleci/src/circleci-config.ts
```diff
@@ -12,7 +12,8 @@ import path from 'path'
 import type { PartialDeep } from 'type-fest'
 import YAML from 'yaml'
 
-const majorOrbVersion = '4'
+const MAJOR_ORB_VERSION = '4'
+const DEFAULT_NODE_VERSION = '16.14-browsers'
```
joelcarr: **thought**: Should we look into programatically using the version the repo has defined in their package.json?

joelcarr: We should do this as part of https://financialtimes.atlassian.net/browse/CPP-1360

PR #433, opened 2023-05-23
Work around incompatibility between Webpack 4 and Node 18

ivomurrell:
# Description

One of the changes going from Node 16 to Node 18 is a [bump](https://nodejs.org/en/blog/release/v17.0.0#openssl-30) to OpenSSL 3. This version of the library drops support for some legacy APIs, one of which is used by the Webpack build process for hashing build artefacts. This results in a [build error](https://app.circleci.com/pipelines/github/Financial-Times/next-static/2481/workflows/b75871a1-8675-4f1a-9cb5-682cfeed3f6a/jobs/10245) like
```
⚠ Error: error:0308010C:digital envelope routines::unsupported
    at new Hash (node:internal/crypto/hash:71:19)
    at Object.createHash (node:crypto:133:10)
    at module.exports (next-static/node_modules/webpack/lib/util/createHash.js:135:53)
    at NormalModule._initBuildHash (next-static/node_modules/webpack/lib/NormalModule.js:417:16)
<etc>
```
It doesn't [look](https://github.com/webpack/webpack/issues/14532) like the Webpack devs are going to backport a fix to Webpack 4. We could update to Webpack 5 to resolve the issue, but I don't think we should ask people to bump their Webpack major version at the same time we're already asking them to bump Node versions. Instead, let's leverage Tool Kit to work around the issue by passing in a [command line argument](https://nodejs.org/docs/latest-v18.x/api/cli.html#--openssl-legacy-provider) to the Node process running Webpack that enables the legacy APIs so that Webpack 4 can continue to function for now.

# Checklist:

### DIFF
--- a/plugins/webpack/test/tasks/webpack.test.ts
+++ b/plugins/webpack/test/tasks/webpack.test.ts
@@ -31,7 +31,7 @@ describe('webpack', () => {
       expect(fork).toBeCalledWith(
         webpackCLIPath,
         ['build', '--color', '--mode=development', '--config=webpack.config.js'],
-        { silent: true }
+        { silent: true, execArgv: expect.arrayContaining([]) }
       )
     })
   })
@@ -45,7 +45,7 @@ describe('webpack', () => {
       expect(fork).toBeCalledWith(
         webpackCLIPath,
         ['build', '--color', '--mode=production', '--config=webpack.config.js'],
-        { silent: true }
+        { silent: true, execArgv: expect.arrayContaining([]) }
       )
     })
   })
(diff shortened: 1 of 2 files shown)


plugins/webpack/test/tasks/webpack.test.ts
```diff
@@ -45,7 +45,7 @@ describe('webpack', () => {
       expect(fork).toBeCalledWith(
         webpackCLIPath,
         ['build', '--color', '--mode=production', '--config=webpack.config.js'],
-        { silent: true }
+        { silent: true, execArgv: expect.arrayContaining([]) }
```
apaleslimghost, line 48: **suggestion**: i think we should add a test for the legacy openssl behaviour, by mocking or overriding `process.allowedNodeEnvironmentFlags` if possible

apaleslimghost (approved):
sensible as a workaround, would be nice to migrate everyone to webpack 5 eventually though 😞

PR #437, opened 2023-05-30
Don't depend on all Node versions for joining jobs in CircleCI configurations

ivomurrell:
# Description

This avoids the [same issue we see in n-gage projects](https://github.com/Financial-Times/n-tracking/pull/119) where depending on all the Node version matrix jobs when 'fanning-in' with a job that does not use a matrix meant that we could not attach the workspace as the `node_modules` directory in the matrix jobs conflicted. The downside now is that the fanned-in job will still run even if a matrix job for an older Node version fails. I don't think CircleCI provides any way to work around this when testing multiple Node versions concurrently (there's no way to select which job to attach a workspace from, for instance).

# Checklist:

### DIFF
--- a/plugins/circleci/src/circleci-config.ts
+++ b/plugins/circleci/src/circleci-config.ts
@@ -109,7 +109,16 @@ export const generateConfigWithJob = (options: JobGeneratorOptions): CircleCISta
         ),
         ...matrixBoilerplate()
       }
-    : { requires: options.requires, ...nightlyBoilerplate }
+    : {
+        // only require the latest Node version of a matrix job in order to
+        // avoid workspace conflicts
+        requires: options.requires.map((dep) => `${dep}-node`),
+        // append the default executor name to the job name so that multiple
+        // non-matrix jobs can be chained one after another without having to
+        // know whether a matrix job precedes them or not
+        name: `${options.name}-node`,
+        ...nightlyBoilerplate
+      }
   const config: CircleCIStatePartial = {
     workflows: {
       'tool-kit': {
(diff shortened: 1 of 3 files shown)


plugins/circleci/src/circleci-config.ts
```diff
@@ -109,7 +109,16 @@ export const generateConfigWithJob = (options: JobGeneratorOptions): CircleCISta
...
+    : {
+        // only require the latest Node version of a matrix job in order to
+        // avoid workspace conflicts
+        requires: options.requires.map((dep) => `${dep}-node`),
+        // append the default executor name to the job name so that multiple
+        // non-matrix jobs can be chained one after another without having to
+        // know whether a matrix job precedes them or not
+        name: `${options.name}-node`,
+        ...nightlyBoilerplate
+      }
```
joelcarr, line 121: **question(non-blocking):** do we choose which version of node is the primary? When we start supporting Node 20, we will want the follow jobs to require the Node 18 build and test jobs.

joelcarr (approved):
Happy with this! Just one non-blocking question which if anything needs to be actioned, can be created as a ticket.

plugins/circleci/src/circleci-config.ts
ivomurrell: yes, whichever is the first version in the `nodeVersion` option will be the default `node` executor. I've tried to make that clear in the [docs](https://github.com/Financial-Times/dotcom-tool-kit/blame/b60dad0df89902ea127a90dcddf036980cf5808e/plugins/circleci/readme.md#L32) too

PR #443, opened 2023-06-05
Fix CircleCI configuration error by never appending executor name to approval job

ivomurrell:
# Description

This keeps the code consistent across jobs that do and do not use Node matrices. The `waiting-for-approval` job cannot take Node versions as a matrix as it is not a job defined in the `dotcom-tool-kit` orb, so we don't append the executor to its name either. This should happen in jobs that don't use multiple executors too so we don't get CircleCI errors about referring to a `waiting-for-approval-node` job that has never been defined.

I thought I'd caught all the bugs introduced by #432 in #439 after testing with the `n-express` project. However, this issue is only found in repositories that use a job that shouldn't be split into a matrix but also needs to wait for approval, i.e., the `deploy-review` job used for Heroku review deployments, so I only encountered it after testing a Heroku app again. I've added a snapshot test as part of the PR in the hopes that it will be easier to catch issues like this earlier on.

# Checklist:

### DIFF
--- a/plugins/circleci/test/circleci-config.test.ts
+++ b/plugins/circleci/test/circleci-config.test.ts
@@ -28,7 +28,7 @@ describe('CircleCI config hook', () => {
     config = generateConfigWithJob({
       name: 'test-job',
       addToNightly: true,
-      requires: ['another-job'],
+      requires: ['waiting-for-approval', 'that-job'],
       splitIntoMatrix: false
     })
   }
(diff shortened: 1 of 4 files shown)


plugins/circleci/test/circleci-config.test.ts
```diff
@@ -191,5 +191,18 @@ describe('CircleCI config hook', () => {
       }
       expect(jobs[3]).toEqual(expect.objectContaining(partialExpectedJob))
     })
+
+    it('should match the snapshot', async () => {
```
joelcarr, line 195: **suggestion:** We should rephrase this test to say they purpose of the snapshot.

joelcarr (approved):

PR #447, opened 2023-06-13
Disable Node 18's native fetch across all plugins

ivomurrell:
# Description

We don't yet want to support Node 18's new fetch implementation. Tool Kit is enabling this by passing the `--no-experimental-fetch` flag to processes forked by the Node and Nodemon tasks. However, plenty of other tasks (e.g., Jest and Mocha) also involve spawning child Node processes and they don't have this flag passed to them. This can cause issues when running tests, for example.

Instead of adding a code snippet to each relevant task to make sure fetch is disabled, let's handle the flag at the top-level by setting it in the global `process.execArgv` object that will be read by all calls to `fork()`.

This presents the issue of how we're going to let users opt out of the behaviour if they legitimately want to use the new fetch. Previously, users would have to set an option for each plugin, but this won't work when we're setting the flags once globally rather than per plugin. It would make more sense to specify an option that will apply globally. So users can now pass global options to a `dotcom-tool-kit` key in their options object. We can add more options in the future, but for now the only option is to enable Node 18's native fetch (off by default).

I've also removed the per plugin options. This is _technically_ a breaking change, but I've [searched](https://github.com/search?q=org%3AFinancial-Times%20allowNativeFetch&type=code) GitHub and no one was using this escape hatch yet so I think we can get away with yanking it out again.

# Checklist:

### DIFF
--- a/core/cli/readme.md
+++ b/core/cli/readme.md
@@ -1,0 +1,13 @@
+# dotcom-tool-kit
+
+_(This README is for the core `dotcom-tool-kit` package that must be installed to use Tool Kit. You can find the documentation for the Tool Kit project itself at https://github.com/Financial-Times/dotcom-tool-kit/blob/main/readme.md.)_
+
+The primary Tool Kit binary that will be invoked to handle all your hooks and tasks.
+
+## Options
+
+There are some global options available that will affect all plugins. All are optional but you can override them with the `dotcom-tool-kit` key.
+
+| Key | Description | Default value |
+|-|-|-|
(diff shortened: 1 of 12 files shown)


core/cli/readme.md
```diff
@@ -0,0 +1,11 @@
+# dotcom-tool-kit
```
apaleslimghost, line 1: **suggestion**: this readme is going to look a bit weird in isolation on npmjs.com. probably better than the current no readme, but. it could link to the toplevel readme or something?

apaleslimghost (approved):
nice!

PR #449, opened 2023-06-20
fix: install Reliability Kit ESLint config and fix errors found

CyntiBinti:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Changes
Agreed with Joel to install on behalf of Platforms team. Found 16 problems (16 errors, 0 warnings)

image

# Checklist:


core/cli/src/plugin.ts
```diff
@@ -142,7 +142,7 @@ export async function loadPlugin(
     ? Promise.resolve({ valid: true, value: undefined })
     : importPlugin(pluginRoot)
 
-  plugin.value.rcFile = await rcFilePromise
+  plugin.value.rcFile = await rcFilePromise // error: Possible race condition: `plugin.value.rcFile` might be assigned based on an outdated state of `plugin` (require-atomic-updates)
```
CyntiBinti, line 145: Not too sure with this error on how best to resolve - any thoughts at all?

apaleslimghost: if i understand the rule correctly, i think it wouldn't error if we did:

```suggestion
  const { value } = plugin
  value.rcFile = await rcFilePromise
```

which seems overly pedantic. in this case it is a pure false positive, since the `plugin` variable isn't from another scope and can't be written to concurrently, so i think disabling it for this line would be appropriate

honestly i don't think this ESLint rule gives us any value at all? i've only seen it return false positives, and it's trying to catch an error caused by a style of code i don't think i've ever seen in Customer Products repos. would we consider removing it from the Reliability Kit config?

CyntiBinti: Ohh fair, I see what you mean about the false positives. Let me raise your point and get back to you on that 😄 -- thank you!

lib/vault/src/index.ts
```diff
@@ -182,6 +182,7 @@ export class VaultEnvVars {
         if (allShared.hasOwnProperty(envVar)) {
           required[envVar] = allShared[envVar]
         }
+        return required[envVar];
```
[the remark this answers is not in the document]
CyntiBinti: Good point - will change this to a forEach loop instead

PR #456, opened 2023-08-11
CPP-925:  Stop force using npm 8 on circleci

joelcarr:
This is no longer needed now all systems use atleast Node v16 which is bundled with npm v8.

### DIFF
--- a/orb/src/jobs/setup.yml
+++ b/orb/src/jobs/setup.yml
@@ -2,6 +2,10 @@ parameters:
   executor:
     default: default
     type: executor
+  npm-version:
+    description: This value is deprecated and no longer used. We now use Node's bundled version of npm.
+    default: ''
+    type: string
 
 executor: << parameters.executor >>
 
--- a/orb/src/jobs/setup.yml
+++ b/orb/src/jobs/setup.yml
@@ -4,13 +4,17 @@ parameters:
     type: executor
   npm-version:
     description: This value is deprecated and no longer used. We now use Node's bundled version of npm.
-    default: ''
+    default: ""
     type: string
 
 executor: << parameters.executor >>
 
 steps:
   - attach-workspace
+  - when:
+      condition: << parameters.npm-version >>
+      steps:
+          - run: sudo npm install -g npm@$PARAM_NPM_VERSION
   - node/install-packages
   - persist-workspace:
       path: .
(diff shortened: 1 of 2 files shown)


orb/src/jobs/setup.yml
```diff
@@ -2,15 +2,11 @@ parameters:
   executor:
     default: default
     type: executor
-  npm-version:
-    default: '8'
-    type: string
```
ivomurrell, line 7: **issue**: deleting this parameter will be a breaking change as any project that was setting it previously will now error as the parameter doesn't exist anymore. this might technically be permissible as no project seems to be setting it currently, but it might be preferable to change the logic to default to `''` and then only run the command if the parameter isn't the empty string. so it will use the bundled npm version by default whilst still allowing users to override it.

joelcarr: I have revieved the value and added a deprecation description.

orb/src/jobs/setup.yml
```diff
@@ -3,14 +3,14 @@ parameters:
...
-    default: '8'
+    description: This value is deprecated and no longer used. We now use Node's bundled version of npm.
+    default: ''
     type: string
 
 executor: << parameters.executor >>
 
 steps:
   - attach-workspace
-  - run: sudo npm install -g npm@<<parameters.npm-version>>
```
ivomurrell, line 13: **suggestion**: this is still a breaking change, let's support the `npm-version` if a user sets it for now
```yaml
  - run:
      command: |-
        if [ -n "$PARAM_NPM_VERSION" ]; then
          sudo npm install -g npm@$PARAM_NPM_VERSION
        fi
      environment:
        PARAM_NPM_VERSION: <<parameters.npm-version>>
```
(inspired by how the CircleCI's `node` [orb](https://circleci.com/developer/orbs/orb/circleci/node#orb-source) handles empty string parameters)

joelcarr: Just found out about the `when` and `unless` properties for CircleCI 🤯 https://circleci.com/docs/configuration-reference/#the-when-step

It means the above code can be done with just:

```yml
  - when:
      condition: << parameters.npm-version >>
      steps:
          - run: sudo npm install -g npm@$PARAM_NPM_VERSION    
```

ivomurrell (approved):
looks great!

PR #457, opened 2023-08-11
feat(orb): default executor to be Node LTS version

joelcarr:
As we are well in the progress of all Customer Products Systems and Packages using Node 18, we should default that our CI runs on LTS (unless stated otherwise).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ references:
       - image: cimg/node:<< parameters.node-version >>
     parameters:
       node-version:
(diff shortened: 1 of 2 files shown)


ivomurrell (changes requested):
**issue**: I don't think it's a good idea to not pin the version here: it means any projects using the `lts` tag will automatically switch to Node 20 in [2.5 months](https://github.com/nodejs/release#release-schedule) from now

joelcarr:
We should control the version we support as Ivo has pointed out.

PR #460, opened 2023-08-29
feat(cypress): pull vault env variables into local hook

cerdfied:
# Description

`next-appetite` currently [extends the CypressLocal task ](https://github.com/Financial-Times/next-appetite/blob/main/toolkit/cypress/index.js)to pull in vault environment variables.

We want to do similar for `cp-content-pipeline` therefore we should have this logic in tool-kit.
# Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,13 +47,13 @@
     },
     "core/cli": {
       "name": "dotcom-tool-kit",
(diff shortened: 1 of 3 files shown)


apaleslimghost (approved):
nice!

PR #462, opened 2023-09-04
chore: store e2e test artifacts in circleci

cerdfied:
# Description

We want to store Cypress artifacts in CircleCi for debugging purposes.

# Checklist:

### DIFF
--- a/orb/src/jobs/e2e-test-review.yml
+++ b/orb/src/jobs/e2e-test-review.yml
@@ -10,3 +10,5 @@ steps:
   - run:
       name: Smoke tests
       command: npx dotcom-tool-kit test:review
+  - store_artifacts:
+      path: /root/project/e2e-test-results
--- a/orb/src/jobs/e2e-test-staging.yml
+++ b/orb/src/jobs/e2e-test-staging.yml
@@ -14,3 +14,5 @@ steps:
       when: always
       name: Scale down staging
       command: npx dotcom-tool-kit teardown:staging
+  - store_artifacts:
+      path: /root/project/e2e-test-results


apaleslimghost (approved):
nice!

**nitpick**: this commit should be `feat:` not `chore:`, so it can be released

orb/src/jobs/e2e-test-review.yml
```diff
@@ -10,3 +10,5 @@ steps:
   - run:
       name: Smoke tests
       command: npx dotcom-tool-kit test:review
+  - store_artifacts:
+      path: /e2e-test-results
```
ivomurrell, line 14: **suggestion**: we should use relative paths from the build directory instead of storing artefacts at the root of the filesystem (I'm not sure we even have permission to create new directories at root?)
```suggestion
      path: e2e-test-results
```

ivomurrell (changes requested):
**todo**: this is a new feature for the orb so should be committed with the `feat` commit type

orb/src/jobs/e2e-test-review.yml
apaleslimghost: this isn't a path to create artefacts in, it's the (existing) path to save files from as artefacts

ivomurrell: I'm assuming we'll have to configure Cypress (and any other e2e tester) to store artefacts in this `e2e-test-results` directory and I don't think we should be configuring it to store the results at `/e2e-test-results`

ivomurrell:
**question**: should we include the change to get Cypress to store artefacts in `e2e-test-results` in this PR too?

orb/src/jobs/e2e-test-review.yml
cerdfied: oh yeah I already store artifacts in cypress/videos so will just set to that

apaleslimghost: because these orb tasks could be used for another e2e test plugin not just Cypress, the path it's saving should be generic, and we should configure Cypress to save videos to that generic path

cerdfied:
> **question**: should we include the change to get Cypress to store artefacts in `e2e-test-results` in this PR too?

have configured it [here](https://github.com/Financial-Times/cp-content-pipeline/pull/516/files#diff-3f62e885350a48939f240b9daf40f6b06d41da7751f2b613238fc7669ac0a1b2R5)

PR #466, opened 2023-09-18
CPP-1695: Serverless deploy prioritise env vars from secrets management service

joelcarr:
# Description

The CI environment was overwriting the values needed for production / other remote environments.

# Checklist:

### DIFF
--- a/plugins/serverless/src/tasks/deploy.ts
+++ b/plugins/serverless/src/tasks/deploy.ts
@@ -35,10 +35,7 @@ export default class ServerlessDeploy extends Task<typeof ServerlessSchema> {
       }
 
       const child = spawn('serverless', args, {
-        env: {
-          ...dopplerEnv,
-          ...process.env
-        }
+        env: dopplerEnv
       })
 
       hookFork(this.logger, 'serverless', child)
--- a/plugins/serverless/src/tasks/provision.ts
+++ b/plugins/serverless/src/tasks/provision.ts
@@ -42,10 +42,7 @@ export default class ServerlessProvision extends Task<typeof ServerlessSchema> {
     }
 
     const child = spawn('serverless', args, {
-      env: {
-        ...dopplerEnv,
-        ...process.env
-      }
+      env: dopplerEnv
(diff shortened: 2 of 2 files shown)


ivomurrell (changes requested):
**suggestion**: [looking](https://github.com/Financial-Times/dotcom-tool-kit/blob/dc016267972f89d0e38aa1979abd07693f831960/plugins/heroku/src/setConfigVars.ts#L29) at the Heroku plugin I think it would be better to ignore `process.env` entirely when deploying. Vault/Doppler is the source of truth for what is configured in Heroku/AWS and the whatever is set in CircleCI should be ignored.

plugins/serverless/src/tasks/deploy.ts
```diff
@@ -36,8 +36,7 @@ export default class ServerlessDeploy extends Task<typeof ServerlessSchema> {
 
       const child = spawn('serverless', args, {
         env: {
-          ...dopplerEnv,
-          ...process.env
+          ...dopplerEnv
         }
```
ivomurrell, line 40: **nitpick**: we don't need to spread the object anymore

```suggestion
        env: dopplerEnv
```

plugins/serverless/src/tasks/deploy.ts
ivomurrell, line 1: **thought**: does the `useVault` option actually make any sense here? we don't seem to provide an analogous option in Heroku and and I don't know what the environment should look like if we were to skip Vault/Doppler...

ivomurrell (approved):
two non blocking comments but fine to merge

plugins/serverless/src/tasks/deploy.ts
joelcarr: We want to get environment variables from the secrets management service to inject into the environment that `serverless.yml` will get the secrets from. Ideally we would use AWS Parameter Store instead meaning we will not need to fetch secrets from the secrets management service but we still need to figure out what that process will look like.

ivomurrell (approved):

PR #472, opened 2023-09-22
CPP-1509 Support new Heroku apps which now include random identifier

ivomurrell:
# Description

Heroku made a [change](https://devcenter.heroku.com/changelog-items/2597) to the naming of the domain apps run on a couple of months ago. Previously, the domain would be based on the name of the app, but now a random 12-character string is appended to make them more unique I guess. This change only affects newly created apps.

We were assuming the previous format when constructing the URL for gtg tests. This meant that gtg tests have started to fail when deploying new apps. The correct domain is actually returned by the [same API](https://devcenter.heroku.com/articles/platform-api-reference#app-info) that we were previously using to check app names/convert to names from an app ID, so let's use that. This works for apps using the old domain naming system as well as the new one.

Tested on `ft-next-static-staging` using the [cursed script](https://github.com/Financial-Times/next-static/tree/main/scripts/test-deployment).

Closes #471.

# Checklist:

### DIFF
--- a/plugins/heroku/src/declarations.d.ts
+++ b/plugins/heroku/src/declarations.d.ts
@@ -57,6 +57,7 @@ declare module 'heroku-client' {
 
   export type HerokuApiResGetGtg = {
     name: string
+    web_url: string | null
   }
 
   export type HerokuApiResPost = {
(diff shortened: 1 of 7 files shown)


joelcarr (approved):

cerdfied (approved):
lgtm!

PR #474, opened 2023-09-28
Skip syncing Doppler secrets to Heroku in favour of official integration

ivomurrell:
# Description

We plan to use the Doppler's official [Heroku integration](https://docs.doppler.com/docs/heroku) so we don't need to upload secrets from Doppler to Heroku here. Also note that the generated config vars (`REGION`, `NODE_ENV`, and `SYSTEM_CODE`) will now be stored in Doppler in [branch configs](https://docs.doppler.com/docs/branch-configs) instead to maintain a single source of truth (and also because it will override any we set ourselves here.)

# Checklist:

### DIFF
--- a/plugins/heroku/test/setConfigVars.test.ts
+++ b/plugins/heroku/test/setConfigVars.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, jest } from '@jest/globals'
 import { setAppConfigVars, setStageConfigVars } from '../src/setConfigVars'
-import { DopplerEnvVars } from '@dotcom-tool-kit/doppler'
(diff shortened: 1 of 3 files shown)


plugins/heroku/test/setConfigVars.test.ts
```diff
@@ -108,4 +113,14 @@ describe('setConfigVars', () => {
   it('resolves if successful', async () => {
     await expect(setAppConfigVars(logger, appName, environment, systemCode)).resolves.not.toThrow()
   })
+
+  it('does not set config vars when Doppler is in use', async () => {
+    // HACK:20230928:IM there's probably a better way to do this but I don't
+    // enjoy playing with Jest mocks
```
cerdfied, line 119: 😆

cerdfied (approved):

PR #476, opened 2023-10-03
Support randomised Heroku subdomains in n-test

ivomurrell:
# Description

Following on from #472, we also need add support the new Heroku app domain names in n-test. We now try to get the `web_url` returned from Heroku in n-test whilst falling back to the old method of using the app name to ensure backwards compatibility (though I can't think of a concrete way it would ever be an issue...)

# Checklist:


cerdfied (approved):

joelcarr (approved):

plugins/n-test/src/tasks/n-test.ts
```diff
@@ -12,7 +12,10 @@ export default class NTest extends Task<typeof SmokeTestSchema> {
 
     // if we've built a review or staging app, test against that, not the app in the config
     if (appState) {
-      this.options.host = `https://${appState.appName}.herokuapp.com`
+      // HACK:20231003:IM keep the old logic of using the app name as the
+      // subdomain to maintain backwards compatibility
+      this.options.host =
+        'url' in appState && appState.url ? appState.url : `https://${appState.appName}.herokuapp.com`
```
joelcarr, line 18: **praise:** this reduces the complexity to [add n-test support for serverless apps](https://financialtimes.atlassian.net/browse/CPP-1630) 🎉

PR #482, opened 2023-10-04
CPP-1664: deprecate pa11y

joelcarr:
# Description

We are deprecating the Pa11y plugin as it is no longer recommended for accessibility testing within Customer Products. This PR adds the ability to run Pa11y locally (as an immediate step to support teams relying on the tool for now) then a deprecation warning along with suggestions for next steps.

# Checklist:

### DIFF
--- a/plugins/pa11y/.toolkitrc.yml
+++ b/plugins/pa11y/.toolkitrc.yml
@@ -2,4 +2,4 @@ plugins:
   - '@dotcom-tool-kit/doppler'
 
 hooks:
-  'test:review': Pa11y
+  'test:local': Pa11y
--- a/plugins/pa11y/src/tasks/pa11y.ts
+++ b/plugins/pa11y/src/tasks/pa11y.ts
@@ -15,9 +15,7 @@ export default class Pa11y extends Task<typeof Pa11ySchema> {
 
     if (localState) {
       process.env.TEST_URL = `https://local.ft.com:${localState.port}`
-    }
-
-    if (reviewState) {
+    } else if (reviewState) {
       process.env.TEST_URL = `https://${reviewState.appName}.herokuapp.com`
     }
 
--- a/plugins/pa11y/src/tasks/pa11y.ts
+++ b/plugins/pa11y/src/tasks/pa11y.ts
@@ -10,7 +10,13 @@ export default class Pa11y extends Task<typeof Pa11ySchema> {
   static description = ''
 
   async run(): Promise<void> {
+    const localState = readState('local')
     const reviewState = readState('review')
+
+    if (localState) {
+      process.env.TEST_URL = `https://local.ft.com:${localState.port}`
+    }
+
     if (reviewState) {
       process.env.TEST_URL = `https://${reviewState.appName}.herokuapp.com`
     }
--- a/plugins/pa11y/test/pa11y.test.ts
+++ b/plugins/pa11y/test/pa11y.test.ts
@@ -21,12 +21,31 @@ jest.mock('@dotcom-tool-kit/logger')
 jest.mock('@dotcom-tool-kit/state')
 
 describe('pa11y', () => {
-  it("sets process.env.TEST_URL if readState('review') is truthy", async () => {
-    jest.mocked(state.readState).mockReturnValue({ appName } as state.ReviewState)
-
+  let MOCK_ENV: 'local' | 'ci' = 'local'
+  beforeEach(() => {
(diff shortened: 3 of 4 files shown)


plugins/pa11y/.toolkitrc.yml
```diff
@@ -2,4 +2,4 @@ plugins:
   - '@dotcom-tool-kit/doppler'
 
 hooks:
-  'test:review': Pa11y
+  'test:local': Pa11y
```
ivomurrell, line 5: **issue**: removing the default link between the `test:review` hook and the `Pa11y` task is a breaking change as it will mean Pa11y will stop running in end to end tests. we should keep the association between the two until we release a new major version
```suggestion
  'test:review': Pa11y
  'test:local': Pa11y
```

plugins/pa11y/src/tasks/pa11y.ts
```diff
@@ -10,7 +10,13 @@ export default class Pa11y extends Task<typeof Pa11ySchema> {
...
 
   async run(): Promise<void> {
+    const localState = readState('local')
     const reviewState = readState('review')
+
+    if (localState) {
+      process.env.TEST_URL = `https://local.ft.com:${localState.port}`
+    }
+
     if (reviewState) {
```
ivomurrell, line 20: **nitpick**: using an `else if` conditional would make it clearer that the `TEST_URL` will be overwritten if `reviewState` is truthy, even if `localState` is also truthy
```suggestion
    } else if (reviewState) {
```

plugins/pa11y/.toolkitrc.yml
joelcarr: The tasks doesn't do anything in `test:review` so I believe we should release this deprecation as a breaking change.

ivomurrell: To clarify: do you mean that Pa11y doesn't work in CI? Why do you say the tasks don't do anything in the `test:review` hook?

joelcarr: Pa11y fails to execute the tests when running on CircleCI however it fails silently and the job will not fail which is misleading people to believe nothing is wrong.

I have updated the commit message to be a breaking change and changed the wording to be clear of the intent.

plugins/pa11y/test/pa11y.test.ts
```diff
@@ -21,12 +21,31 @@ jest.mock('@dotcom-tool-kit/logger')
 jest.mock('@dotcom-tool-kit/state')
 
 describe('pa11y', () => {
-  it("sets process.env.TEST_URL if readState('review') is truthy", async () => {
-    jest.mocked(state.readState).mockReturnValue({ appName } as state.ReviewState)
-
+  let MOCK_ENV: 'local' | 'ci' = 'local'
```
ivomurrell, line 24: **thought**: jest is so annoying lol

plugins/pa11y/test/pa11y.test.ts
```diff
@@ -21,12 +21,31 @@ jest.mock('@dotcom-tool-kit/logger')
...
+  beforeEach(() => {
+    jest.mocked(state.readState).mockImplementation((stateType) => {
+      let returnState
+      if (stateType === 'local' && MOCK_ENV === 'local') {
+        returnState = { port: 5050 } as state.LocalState
+      }
+      if (stateType === 'review' && MOCK_ENV === 'ci') {
```
ivomurrell, line 31: **nitpick**: this could also use `else if` as they're mutually exclusive 😄

```suggestion
      } else if (stateType === 'review' && MOCK_ENV === 'ci') {
```

plugins/pa11y/test/pa11y.test.ts
ivomurrell, line 1: **nitpick**: the test fixes are stuffed into an unrelated `refactor` commit

ivomurrell (approved):
LGTM!

plugins/pa11y/test/pa11y.test.ts
joelcarr: The refactor caused the test to break but also caused the need for the test to be accurate 😂 Before the the `else if` implementation, the test was like "yep there is a local state, let me right TEST_URL. Oh there is also a review state, let me overwrite TEST_URL". I am happy to split it out.

ivomurrell: ah lol, I understand now, no it makes sense to keep it all together then!

PR #483, opened 2023-10-06
CPP-1666 Try upper-cased secret names for AWS credentials to aid Doppler migration

ivomurrell:
# Description

This is to support the Doppler migration, as Doppler doesn't support secret names with lowercase characters.

# Checklist:

### DIFF
--- a/plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
+++ b/plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
@@ -76,16 +76,24 @@ export default class UploadAssetsToS3 extends Task<typeof UploadAssetsToS3Schema
       throw new ToolKitError(`no files found at the provided directory: ${options.directory}`)
     }
 
+    // HACK:20231006:IM Doppler doesn't support secrets with lowercase
+    // characters so let's check if the provided AWS environment variable name
+    // is available in all caps as will be the case when running our migration
+    // script.
+    const checkUppercaseName = (envName: string): string | undefined => {
+      return process.env[envName] ?? process.env[envName.toUpperCase()]
+    }
     const s3 = new S3Client({
       region: options.region,
-      // fallback to default value for accessKeyId if neither accessKeyIdEnvVar or accessKeyId have been provided as options
+      // will fallback to default value for accessKeyId if neither
+      // accessKeyIdEnvVar nor accessKeyId have been provided as options
       credentials: {
         accessKeyId:
-          /* eslint-disable-next-line */
-          process.env[options.accessKeyIdEnvVar ?? options.accessKeyId]!,
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          checkUppercaseName(options.accessKeyIdEnvVar ?? options.accessKeyId)!,
         secretAccessKey:
-          /* eslint-disable-next-line */
-          process.env[options.secretAccessKeyEnvVar ?? options.secretAccessKey]!
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          checkUppercaseName(options.secretAccessKeyEnvVar ?? options.secretAccessKey)!
       }
     })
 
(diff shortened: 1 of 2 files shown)


plugins/upload-assets-to-s3/src/tasks/upload-assets-to-s3.ts
```diff
@@ -76,16 +76,24 @@ export default class UploadAssetsToS3 extends Task<typeof UploadAssetsToS3Schema
       throw new ToolKitError(`no files found at the provided directory: ${options.directory}`)
     }
 
+    // HACK:20231006:IM Doppler doesn't support secrets with lowercase
+    // characters so let's check if the provided AWS environment variable name
+    // is available in all caps as will be the case when running our migration
+    // script.
+    const checkUppercaseName = (envName: string): string | undefined => {
+      return process.env[envName] ?? process.env[envName.toUpperCase()]
```
joelcarr, line 84: **question:** It has taken me a while to figure out what is happening here. Commenting some test scenarios here for a confidence check.

```
SCENARIO uploading assets for a project that uses Vault
GIVEN the project Tool Kit config has defined `accessKeyIdEnvVar` as `aws_access_hashed_assets` for the plugin
AND there is an environment variable set `aws_access_hashed_assets` set by next-vault-sync
WHEN uploadAssetsToS3 executes
THEN the environment variable `aws_access_hashed_assets` will be used
```

```
SCENARIO uploading assets for a project that uses Doppler
GIVEN the project Tool Kit config has defined `accessKeyIdEnvVar` as `aws_access_hashed_assets` for the plugin
AND there is an environment variable set `AWS_ACCESS_HASHED_ASSETS` set by Doppler
WHEN uploadAssetsToS3 executes
THEN the environment variable `AWS_ACCESS_HASHED_ASSETS` will be used
```

Is this along the right lines?

joelcarr: **suggestion:** I see there is tests setup for this plugin, would be great to define some tests for this

ivomurrell: both your scenarios are correct. more generically, the scenario is almost all projects will be checking the environment variable `aws_access_hashed_assets` as that's the [default value](https://github.com/Financial-Times/dotcom-tool-kit/blob/6c9ee404250257ee528e17c69bffc5dca29d16d2/lib/types/src/schema/upload-assets-to-s3.ts#L6) if not overridden in a user's `.toolkitrc.yml`. This works fine with Vault, but Doppler doesn't support lowercase characters in secret names so we store the secret as `AWS_ACCESS_HASHED_ASSETS`. the code checks for the environment variable name as given first; by default the name will be `aws_access_hashed_assets`, which will work if Vault is used but not Doppler. if the environment variable is `undefined` (as it will be with Doppler) we then fallback to checking the `envName.toUpperCase()`, so `"aws_access_hashed_assets".toUpperCase()`, so `AWS_ACCESS_HASHED_ASSETS`. this will succeed with Doppler.

plugins/upload-assets-to-s3/test/tasks/upload-assets-to-s3.test.ts
```diff
@@ -105,4 +105,22 @@ describe('upload-assets-to-s3', () => {
...
+  // HACK:20231006:IM make sure hack to support Doppler migration works
+  it('should fallback to uppercase environment variable', async () => {
+    const task = new UploadAssetsToS3(logger, {
+      ...defaults,
+      directory: testDirectory
+    })
+    // must use delete to ensure an environment variable is undefined, setting
+    // 'undefined' will just stringify the word
+    delete process.env.aws_access_hashed_assets
+    process.env.AWS_ACCESS_HASHED_ASSETS = '1234'
```
joelcarr, line 118: **praise:** This is a clear test which helps explains what is happening! Thanks for adding this 🙌

joelcarr (approved):

PR #486, opened 2023-10-09
Strip ANSI escape codes that mess with terminal state, starting with RIS

ivomurrell:
# Description

The [`RIS`](https://en.wikipedia.org/wiki/ANSI_escape_code#Fs_Escape_sequences) escape code is [used](https://github.com/microsoft/TypeScript/blob/61a96b1641abe24c4adc3633eb936df89eb991f2/src/compiler/sys.ts#L1604C39-L1604C44) by TypeScript's watch mode to clear the terminal before printing its watch logs. Rather than stripping all the (mostly useful) ANSI escape codes, let's just remove this one that can interfere when Tool Kit is printing logs from multiple tasks.

# Checklist:

### DIFF
--- a/lib/logger/src/helpers.ts
+++ b/lib/logger/src/helpers.ts
@@ -12,6 +12,11 @@ const whitespaceRegex = /\s|\n/g
 const startRegex = new RegExp(`^(?:(?:${ansiRegexText})|\s|\n)+`)
 const endRegex = new RegExp(`(?:(?:${ansiRegexText})|\s|\n)+$`)
 
+// RIS escape code to effectively do a clear (^L) on the terminal. TypeScript's
+// watch mode does this and it's annoying to have the logs shunted around when
+// tracking multiple tasks.
+const ansiReset = /\x1Bc/g
+
 // Trim whitespace whilst preserving ANSI escape codes
 function ansiTrim(message: string): string {
   let start = 0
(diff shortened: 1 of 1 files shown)


apaleslimghost (approved):

apaleslimghost:
**thought**: `\r` could potentially mess up terminal output too. it's mainly used for drawing progress bars etc though, which we probably don't want to mess with. most things will only draw a progress bar if they detect that their stdout/stderr is a TTY; do our subprocesses think they're in a TTY?

ivomurrell:
> **thought**: `\r` could potentially mess up terminal output too. it's mainly used for drawing progress bars etc though, which we probably don't want to mess with. most things will only draw a progress bar if they detect that their stdout/stderr is a TTY; do our subprocesses think they're in a TTY?

they do not; `stdout` is piped not inherited so that we can map our own prefixes onto the logs. might be worth seeing if there's a way around this at some point because it currently means we tend to force `--colors` in tasks which can be problematic when we run Tool Kit in a terminal that doesn't support the requisite escape codes.

but yeah, I think we should leave `\r` untouched for now until we see it being disruptive. I can imagine it exhibiting weird behaviour if logs are interleaving between two tasks but I haven't noticed that in practice yet

apaleslimghost:
> we tend to force `--colors`

we could use https://www.npmjs.com/package/supports-color/v/9.3.0 to check if we should do that

PR #491, opened 2023-10-17
fix: add newline after ruler in error output

apaleslimghost:
idk how long this has been broken but it looks horrible


ivomurrell (approved):
it's probably been broken since https://github.com/Financial-Times/dotcom-tool-kit/commit/368e52804043f2caa67f1cf9193d09194c5d3c15, thanks for the fix!

PR #497, opened 2023-10-24
CPP-1715 Skip syncing Heroku config variables if Doppler migration flagged in Vault

ivomurrell:
# Description

After [testing](https://app.circleci.com/pipelines/github/Financial-Times/next-static/3072/workflows/0e224b36-e913-4ab2-b045-60d4bd4c93eb/jobs/11497?invite=true#step-103-1058_39) with next-static, I've now realised the approach we were employing to support both Vault and Doppler in the `heroku` plugin was fundamentally flawed. We should ([_should_](https://github.com/Financial-Times/next-static/tree/main/scripts/test-deployment)) always be deploying to Heroku from CircleCI, but we explicitly don't want to give CircleCI Doppler access so we can limit fallout from a partial leak of CircleCI secrets. This means that the check for whether a Doppler project exists for the deploying app will always fail because we don't have auth, and therefore we'll always overwrite the Heroku configuration variables with secrets from Vault. So instead of checking for a Doppler project when deploying a Heroku app, let's make use of the same logic next-vault-sync [uses](https://github.com/Financial-Times/next-vault-sync/pull/227) and check for a `MIGRATED_TO_DOPPLER` secret in Vault.

# Checklist:


joelcarr (approved):
Much nicer 👍

PR #501, opened 2023-10-27
CPP-1722: fix npm publish task

joelcarr:
# Description

npm has started rejecting packages that attempt to publish with a different version defined in its' package.json. Turns out our plugin attempted to set the version in package.json but it wasn't saved. Packages have been publishing with a version number in package.json that differs to the version we state on npm.
 
This fix removes the use of the internal npm libraries and replaces it with using the cli directly.

# Checklist:

### DIFF
--- a/plugins/npm/src/tasks/npm-publish.ts
+++ b/plugins/npm/src/tasks/npm-publish.ts
@@ -7,24 +7,28 @@ import pack from 'libnpmpack'
 import { publish } from 'libnpmpublish'
 import { styles } from '@dotcom-tool-kit/logger'
 import tar from 'tar'
-import { PassThrough as PassThroughStream } from 'stream';
+import { PassThrough as PassThroughStream } from 'stream'
 
-type TagType = "prerelease" | "latest"
+type TagType = 'prerelease' | 'latest'
 
 export default class NpmPublish extends Task {
   static description = ''
 
   getNpmTag(tag: string): TagType {
-    if(!tag) {
-        throw new ToolKitError('CIRCLE_TAG environment variable not found. Make sure you are running this on a release version!')
+    if (!tag) {
+      throw new ToolKitError(
+        'CIRCLE_TAG environment variable not found. Make sure you are running this on a release version!'
+      )
     }
-    if(prereleaseRegex.test(tag)) {
+    if (prereleaseRegex.test(tag)) {
       return 'prerelease'
     }
-    if(releaseRegex.test(tag)) {
+    if (releaseRegex.test(tag)) {
       return 'latest'
     }
-    throw new ToolKitError(`CIRCLE_TAG does not match regex ${semVerRegex}. Configure your release version to match the regex eg. v1.2.3-beta.8`)
+    throw new ToolKitError(
+      `CIRCLE_TAG does not match regex ${semVerRegex}. Configure your release version to match the regex eg. v1.2.3-beta.8`
+    )
   }
 
   async listPackedFiles(tarball: Buffer): Promise<void> {
@@ -32,7 +36,7 @@ export default class NpmPublish extends Task {
 
     new PassThroughStream()
       .end(tarball)
-      .pipe(tar.t({onentry: entry => this.logger.info(`- ${styles.filepath(entry.header.path)}`)}))
+      .pipe(tar.t({ onentry: (entry) => this.logger.info(`- ${styles.filepath(entry.header.path)}`) }))
   }
 
   async run(): Promise<void> {
@@ -43,10 +47,8 @@ export default class NpmPublish extends Task {
 
     const ci = readState('ci')
 
-    if(!ci) {
-      throw new ToolKitError(
-        `Could not find state for ci, check that you are running this task on circleci`
-      )
+    if (!ci) {
+      throw new ToolKitError(`Could not find state for ci, check that you are running this task on circleci`)
     }
 
     const tag = ci.tag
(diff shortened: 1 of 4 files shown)


plugins/npm/test/npm-publish.test.ts
```diff
@@ -3,29 +3,24 @@ import NpmPublish from '../src/tasks/npm-publish'
...
-  return {
-    publish: jest.fn(() => Promise.resolve())
-  }
-})
+jest.mock('node:child_process', () => ({
+  spawn: jest.fn(() => Promise.resolve())
+}))
+jest.mock('@dotcom-tool-kit/logger', () => ({
+  waitOnExit: jest.fn(() => Promise.resolve())
+}))
```
ivomurrell, line 17: **suggestion**: this can be simplified to use jest's auto mocking

```suggestion
jest.mock('node:child_process')
jest.mock('@dotcom-tool-kit/logger')
```

plugins/npm/src/tasks/npm-publish.ts
```diff
@@ -1,52 +1,53 @@
...
-  async listPackedFiles(tarball: Buffer): Promise<void> {
-    this.logger.info('packed files:')
-
-    new PassThroughStream()
-      .end(tarball)
-      .pipe(tar.t({onentry: entry => this.logger.info(`- ${styles.filepath(entry.header.path)}`)}))
+  async executeNpmTask(npmTask: 'version' | 'publish', options: string[]): Promise<void> {
+    try {
+      this.logger.verbose(`running \`npm ${npmTask} ${options.join(' ')}\``)
+      const task = spawn('npm', [npmTask, ...options])
```
ivomurrell, line 33: **todo**: this process should be hooked so we can decorate logs with context

```suggestion
      const task = spawn('npm', [npmTask, ...options])
      hookFork(this.logger, `npm ${task}`, task)
```

ivomurrell (approved):
**praise**: this must have been difficult to reason about why it was failing!

**thought**: this seems to have been the behaviour for a [month](https://github.blog/changelog/2023-09-27-block-npm-package-publishes-when-names-and-versions-dont-match-between-manifest-and-tarball-package-json/)

**thought**: this all seems to be happening because we don't want to have to make a commit to bump the `package.json` version ourselves before making a git release... maybe we should just encourage people to do that lol

**thought**: I don't have any major objections to this approach to fixing the bug though I will note that we could also keep using the npm libraries (and their stable APIs, I don't think it's fair to classify them as 'internal' as that usually implies instability) and modify the `package.json` in a couple of lines of `fs.readFile + JSON.parse` and `JSON.stringify + fs.writeFile`. I'd recommend waiting for @apaleslimghost to double-check this next week as she may have more insight into why the libraries were chosen instead of invoking the shell, but I don't want this to be blocking.

plugins/npm/src/tasks/npm-publish.ts
```diff
@@ -63,20 +62,8 @@ export default class NpmPublish extends Task {
...
-    await this.listPackedFiles(tarball)
-
-    await publish(manifest, tarball, {
-      access: 'public',
-      defaultTag: npmTag,
-      forceAuth: {
-        token: process.env.NPM_AUTH_TOKEN
-      }
-    })
+    await this.executeNpmTask('version', [tag])
```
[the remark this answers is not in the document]
joelcarr: We do not do any git changes within our CircleCI tasks. This is the method we use for repos like dotcom-page-kit and have not run into this issue.

plugins/npm/src/tasks/npm-publish.ts
```diff
@@ -1,52 +1,54 @@
+import { spawn } from 'node:child_process'
 import { ToolKitError } from '@dotcom-tool-kit/error'
 import { Task } from '@dotcom-tool-kit/types'
 import { semVerRegex, prereleaseRegex, releaseRegex } from '@dotcom-tool-kit/types/lib/npm'
-import pacote from 'pacote'
 import { readState } from '@dotcom-tool-kit/state'
-import pack from 'libnpmpack'
-import { publish } from 'libnpmpublish'
-import { styles } from '@dotcom-tool-kit/logger'
-import tar from 'tar'
```
apaleslimghost, line 9: **suggestion**: imo this change is completely unrelated to the issue that's being fixed, and so should be done in its own PR. these libraries aren't "internal", they're their own public packages with a documented interface. i used the lower-level libraries rather than wrapping the CLI so we'd have more control over the error handling and logging. i'm not completely against this change, i just don't think it should be bundled with this bug fix.

joelcarr: To clarify, is the change being referred to moving away from libraries such as `libnpmpublish` to using the npm CLI directly? If so, I am not sure how to split this away from the current PR. The code changes to move to using the CLI will be the same as fixing this bug. I can change the commit message to describe the change as moving away from `libnpm*` packages rather than fixing the bug?

I understand that `libnpmpack` and others are published public packages, my reference to them being internal is because the npm CLI internals use the packages.

apaleslimghost: you could have fixed the bug separately via Ivo's suggestion of directly writing the version to `package.json`. moving away from these libraries is a much bigger change to how the task works, which makes me nervous.

joelcarr: As discussed on slack - we are okay to merge this as is for now.

ivomurrell: > As discussed on slack - we are okay to merge this as is for now.

hey, in the future could we try and keep code review discussion within the PR itself rather than moving to Slack DMs? I acknowledge it's slightly harder to hold a back-and-forth discussion in a commenting format rather than an IRC format but I think it's important to be able to be able to easily refer back to why these decisions were made. as it stands, merging this PR which includes various unrelated changes appears to have resulted in many followup bug fix PRs being required, and it's frustrating that I don't have access to the rationale for why this was ultimately merged.

PR #504, opened 2023-11-02
chore: remove and gitignore the sandbox

apaleslimghost:
keeping the sandbox up to date is tedious, and it's less useful now Tool Kit is no longer in the initial development phase.

this PR removes the sandbox contents from the repo, keeping the folder but gitignoring everything except the readme. this lets you create your own sandbox for local testing in that folder, without the overhead of managing its changes when switching between branches, rebasing, and merging, which makes testing things (especially across multiple branches) much easier.


joelcarr (approved):
Thanks for sorting this Kara!

PR #508, opened 2023-11-07
feat(orb): setup npm auth for publish job

joelcarr:
# Description

Now the we use the npm CLI for #501, we need let the CLI known of the auth token. This method was taken from https://github.com/Financial-Times/dotcom-page-kit/blob/6f77ab771c8a915dc58e4535ad8a3735dc57cf67/.circleci/config.yml#L137-L140

# Checklist:

### DIFF
--- a/orb/src/jobs/publish-tag.yml
+++ b/orb/src/jobs/publish-tag.yml
@@ -8,5 +8,9 @@ executor: << parameters.executor >>
 steps:
   - attach-workspace
   - run:
+      name: Set npm auth token
+      command: echo "//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}" >
+        ${HOME}/.npmrc
+  - run:
       name: Publish to npm
       command: npx dotcom-tool-kit publish:tag
--- a/orb/src/jobs/publish.yml
+++ b/orb/src/jobs/publish.yml
@@ -8,5 +8,9 @@ executor: << parameters.executor >>
 steps:
   - attach-workspace
   - run:
+      name: Set npm auth token
+      command: echo "//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}" >
+        ${HOME}/.npmrc
+  - run:
       name: Publish to npm
       command: npx dotcom-tool-kit publish:tag


apaleslimghost (commented):
not super keen on this approach as it ties the `NpmPublish` task to the orb; it could be run manually, or from a custom CircleCI workflow that doesn't use the orb, or from another CI platform. at the very least, we should document this requirement.

an alternative would be to require projects to create and commit a project-specific `.npmrc`, which can reference the environment variable. this is the approach [recommended by the npm docs](https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow#create-and-check-in-a-project-specific-npmrc-file) (in this case for using private packages from CI, but it's the same principle). in the future (once hooks are decoupled from commands) we could have this `.npmrc` created by a hook.

joelcarr:
The `NPM_AUTH_TOKEN` is dependent on the CircleCI environment as it is stored as a CircleCI context. The `NpmPublish` task is dependent on the environment to be set up correctly even before #501.

> it could be run manually, or from a custom CircleCI workflow that doesn't use the orb, or from another CI platform

Tool Kit is setting up the ideal path for the Custom Products publish a package lifecycle event. Adding this auth method to the orb does not prevent our consumers to diverge from this, they will have to follow the npm docs to set up the environment correctly.

Adding the requirement for a consuming project to use a `.npmrc` will be a breaking change. This will also lead to the need of additional documentation for setting it up, why it is there, explaining not to change it, etc. I wish there was a way to just do `npm publish --auth $NPM_AUTH_TOKEN`.

pilargarsev (approved):

ivomurrell:
> I wish there was a way to just do `npm publish --auth $NPM_AUTH_TOKEN`.

The way to do this is to use the JS API npm exports via `libnpmpublish` which provides the [`token`](https://www.npmjs.com/package/libnpmpublish#opts) and [`forceAuth`](https://www.npmjs.com/package/npm-registry-fetch#fetch-opts) options. This is how passing the token was handled before #501; is there a reason why we can't revert back to that approach to avoid having to use `.npmrc`?

joelcarr:
As discussed in Slack, we will be reverting to the original method and finding a fix for that.

PR #513, opened 2023-11-14
WIP: Lazy plugins

apaleslimghost:
this PR implements lazy loading of plugin entrypoints. this is a breaking change, as the plugin interface has changed. further work to make plugins even lazier incoming. this work is the basis of the hook rethink.


core/cli/src/index.ts
```diff
@@ -27,6 +36,23 @@ export const shouldDisableNativeFetch = (): boolean => {
...
+              taskName,
+              new tasks[taskName](logger, taskName, getOptions(pluginId as OptionKey) ?? {})
+            ])
+          )
+        )
+      )
+    ),
+    Object.fromEntries
+  )
+
```
cerdfied, line 55: **suggestion:** these calls are very nested :O is there a way to split this into steps with named variables? I think it would be a lot easier to read

cerdfied (approved):
ok have gone through this line by line and it makes sense to me!

**thought:** have spoken to Kara about this, but worth mentioning here that it took me some time to piece together the flow of loading of config / hooks / plugins / tasks. I think at some point soon, we should rethink the naming of various functions to be a little more explicit eg. `checkInstall`, `check`, `install`. I think it would be a quick win and make reviewing pull requests / debugging a lot easier!

PR #515, opened 2023-11-14
fix conflict message

apaleslimghost:
fixes a bug in the conflict message introduced by the lazy plugins refactor

### DIFF
--- a/core/cli/src/config.ts
+++ b/core/cli/src/config.ts
@@ -6,7 +6,7 @@ import type { Logger } from 'winston'
 
 import type { HookTask } from './hook'
 import { RawPluginModule, importPlugin, loadPlugin, resolvePlugin, validatePluginHooks } from './plugin'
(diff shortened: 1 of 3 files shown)


core/cli/src/config.ts
```diff
@@ -159,8 +159,8 @@ export function validateConfig(config: ValidPluginsConfig, logger: Logger): Vali
   const validConfig = config as ValidConfig
 
   const hookTaskConflicts = findConflicts(Object.values(config.hookTasks))
-  const hookConflicts = findConflicts(Object.values(config.hooks))
-  const taskConflicts = findConflicts(Object.values(config.tasks))
+  const hookConflicts = findConflictingEntries(config.hooks)
+  const taskConflicts = findConflictingEntries(config.tasks)
   const optionConflicts = findConflicts(Object.values(config.options))
```
joelcarr, line 164: **suggestion(non-blocking):** would be great to see some tests for this logic in the future.

PR #518, opened 2023-11-14
refactor: rename main options schema to plugins

apaleslimghost:
groundwork for `hook-options`, which will introduce another type of options and schemas for them


joelcarr (approved):
🔌 📥

package-lock.json
apaleslimghost, line 1: i have no idea where all these package-lock changes have come from 😭

PR #520, opened 2023-11-15
Major release of Tool Kit

apaleslimghost:
~tracking PR for the changes in the next major version. this branch has the same protections as `main`, so it can't be pushed to directly; instead, raise a PR with `next` as the base~

This is now ready to review! This PR includes bundles many breaking changes to be released together in a new major version of Tool Kit. Some of the biggest changes include:

- Optimisation of the loading of plugins, dramatically reducing the startup time of the CLI
- Hooks are now split out into _commands_ and _install hooks_ in order to clarify the two things hooks were previously responsible for – installing config and grouping tasks
- Options can now be set for specific tasks and install hooks exported by plugins rather than all being plugin-wide
- CircleCI and PackageJson install hooks are now customised via options instead of JavaScript subclasses
- Node 20 is now supported and Node 16 is no longer supported


ivomurrell:
I think this is now ready to release. We still want to refine error handling and documentation, but I think we've landed all the breaking changes now and it's already being used in a few maintain and optimise repositories. I don't think we need to review each individual commit because they're all from PRs we've already approved, so this instead calls for a sign-off on releasing (and double checking that we're happy with the conventional commits).

ivomurrell (approved):

PR #521, opened 2023-11-15
chore: update package-lock on next

apaleslimghost:
i have no idea

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,7 @@
     },
     "core/cli": {
       "name": "dotcom-tool-kit",
-      "version": "3.3.6",
+      "version": "3.3.7",
       "license": "MIT",
       "dependencies": {
         "@dotcom-tool-kit/error": "^3.1.0",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -1087,6 +1086,746 @@
         "npm": "7.x || 8.x || 9.x"
       }
     },
+    "core/sandbox/node_modules/@dotcom-tool-kit/backend-heroku-app": {
```
ivomurrell, line 1089: **issue**: why has the sandbox returned 😨

joelcarr:
There are changes in here for `core/sandbox/node_modules/*`. It looks like this branch does not have the sandbox like with `main` so maybe it is cached in some way?

package-lock.json
apaleslimghost: ohhhhhhh. the sandbox is still gitignored, but i still have a `package.json` in that folder which is being picked up in the install. that's annoying. i can't think of a way to skip that without removing it from the workspace which kind of defeats the point of having it gitignored in the monorepo? we ignore these changes though

ivomurrell: could we exclude it in the `workspaces` value in the root `package.json`?

apaleslimghost: i _think_ that would prevent it from picking up other workspace dependencies. we could use file version specifiers for those though?

PR #522, opened 2023-11-15
refactor: use an EntryPoint object for tasks and hooks in the config

apaleslimghost:
instead of just strings, which can easily be confused for strings from somewhere else.

in the very near future the entry point is not just going to be the plugin root; this refactor makes that easier.


cerdfied (approved):
nice, definitely clearer this way to understand the structure of hooks / tasks config!

PR #527, opened 2023-11-17
Even lazier plugins

apaleslimghost:
instead of plugins having a single entry point and doing `pluginModule.tasks[taskName]` when we just need a single thing, have plugins declare what the entry points for tasks and hooks are in their `.toolkitrc.yml`, and just import that. 

can't test this fully yet as i deleted all the hooks but it does at least correctly load tasks.

- [x] get the tests passing (or at least skip them)


cerdfied (approved):
nice! makes sense to me 👍

core/cli/bin/run
```diff
@@ -15,6 +15,9 @@ async function main() {
     } else if (argv.listPlugins) {
       const { listPlugins } = require('../lib')
       await listPlugins(rootLogger)
+    } else if (argv.printConfig) {
+      const { printConfig } = require('../lib')
+      await printConfig(rootLogger)
```
ivomurrell, line 20: **praise**: yeeees

core/cli/test/plugin.test.ts
ivomurrell, line 1: **thought**: maybe we should make a test ensuring `.toolkitrc.yml`s are well-formed instead?

ivomurrell (approved):

PR #532, opened 2023-11-21
move conflict types and handlers to types package

apaleslimghost:


lib/types/src/conflict.ts
```diff
@@ -1,4 +1,4 @@
-import type { Plugin } from '@dotcom-tool-kit/types'
+import type { Plugin } from './'
```
ivomurrell, line 1: **nitpick**: I hate this

```suggestion
import type { Plugin } from './index'
```

core/cli/src/plugin.ts
```diff
@@ -13,7 +13,7 @@ import {
 import resolveFrom from 'resolve-from'
 import type { Logger } from 'winston'
 import { EntryPoint, PluginOptions, RawConfig, ValidPluginsConfig } from './config'
-import { Conflict, isConflict } from './conflict'
+import { Conflict, isConflict } from '@dotcom-tool-kit/types/src/conflict'
```
joelcarr, line 16: **question:** do we want to import from a compiled version of the file rather than from `src`?

lib/types/src/conflict.ts
joelcarr: Please could you elaborate @ivomurrell?

ivomurrell: I don't think it's particularly readable

ivomurrell: I thought it was self-evidently unreadable. If people disagree I'm fine to keep it as it's just a nitpick

joelcarr: I do prefer to be verbose and prefer the change you have suggested. @apaleslimghost are you okay with committing the requested change?

ivomurrell (approved):

PR #549, opened 2023-12-04
Clear config cache to avoid repeated conflict errors during installation in migration script

ivomurrell:
# Description

The cosmiconfig cache has been a [thorn in the side](https://github.com/Financial-Times/dotcom-tool-kit/commit/6d032f4ecae5c532aa08b5105df08339d5636cd1) of the migration script for a while. In this case, we need to clear the cache again after fixing conflicting tasks, as otherwise the explicit ordering of tasks added to the `.toolkitrc.yml` is not picked up. 

# Checklist:


apaleslimghost (approved):
doh, good find

PR #550, opened 2023-12-04
refactor Validated to use chainable methods not nested function calls

apaleslimghost:
easier to read imo


ivomurrell (approved):
much more readable!

PR #551, opened 2023-12-05
PackageJson hook conflict resolution logic

apaleslimghost:


cerdfied (commented):
**praise:** make sense to me, these tests have actually really helped clarify merge behaviour!

joelcarr (commented):
These tests make sense to me and relate to what we have previously discussed.

**question:** I feel like we may have discussed this but can't quite remember - would having a hook per editable property of the file (ie one for `scripts` and one for `lint-staged`) reduce the complexity of managing the conflict resolution across the file?

joelcarr:
From your example I don't believe we are on the same page. Using an example from the test in this PR, I was thinking of something along the lines of:

```
hooks:
  - PackageJsonScripts:
      test: 'test:local',
      build: 'build:local'
  - PackageJsonAnother:
      field: something:else
```

Which once each hook has been actioned, we will expect to output:

```
{
  name: "next-project",
  scripts: {
      test: 'test:local',
      build: 'build:local'
  },
  …
  another: {
    field: 'something: else'
  }
}
```

My thinking is that we are targeting the scope of the changes to reduce the potential conflicts / changes in the API. The hope is that when it comes to the CircleCI config hook(s), we can do something similar where we don't have to concern ourselves with how the whole file is structured as we will look to target specific parts. Does that make sense?

plugins/package-json-hook/src/package-json-helper.ts
```diff
@@ -22,13 +25,125 @@ interface PackageJsonState {
...
+    )[]
+  ): (HookInstallation<z.output<typeof PackageJsonSchema>> | Conflict<HookInstallation>)[] {
+    let mergeable: HookInstallation<z.output<typeof PackageJsonSchema>>[] = []
+    let unmergeable: HookInstallation<z.output<typeof PackageJsonSchema>>[] = []
+
+    for (const installation of childInstallations) {
+      if (isConflict(installation)) {
+        unmergeable.push(...installation.conflicting)
+      } else {
+        ;[mergeable, unmergeable] = partitionInstallations(installation, mergeable, unmergeable)
```
ivomurrell, line 111: **thought**:

[Prettier waiting room]

Me: [chanting] asi, asi-

Other patients: asi, ASI

Secretary: [pounding her clipboard] ASI, ASI, ASI!

ivomurrell (approved):

plugins/package-json-hook/src/package-json-helper.ts
apaleslimghost: 🦷 🦷 🦷 🦷 🦷 🦷 🦷

PR #554, opened 2023-12-15
CPP-1505 Add option in migration script to add .toolkitstate directory to gitignore

ivomurrell:
# Description

We don't want people committing the `.toolkitstate` directory, so let's add make sure it's ignored by git in the migration script. 

# Checklist:

### DIFF
--- a/core/create/src/prompts/main.ts
+++ b/core/create/src/prompts/main.ts
@@ -1,10 +1,16 @@
 import { styles } from '@dotcom-tool-kit/logger'
 import type { PackageJson } from 'type-fest'
-import { existsSync } from 'fs'
+import { existsSync, promises as fs } from 'fs'
 import prompt from 'prompts'
 import { BizOpsSystem } from '../bizOps'
 
-type PromptNames = 'preset' | 'additional' | 'addEslintConfig' | 'deleteConfig' | 'uninstall'
+type PromptNames =
+  | 'preset'
+  | 'additional'
+  | 'addEslintConfig'
+  | 'deleteConfig'
+  | 'ignoreToolKitState'
+  | 'uninstall'
 
 export interface MainParams {
   bizOpsSystem?: BizOpsSystem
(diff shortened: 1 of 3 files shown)


core/create/src/prompts/main.ts
```diff
@@ -99,6 +113,16 @@ export default async ({
...
+      {
+        name: 'ignoreToolKitState',
+        type: missingStateInGitignore ? 'confirm' : null,
+        initial: true,
+        message: `Would you like the ${styles.filepath(
+          '.toolkitstate/'
+        )} directory to be added to your ${styles.filepath(
+          '.gitignore'
+        )}? Tool Kit stores some state files that shouldn't be committed to your repository.`
+      },
```
joelcarr, line 125: **opinion:** I believe we shouldn't give the option and perform this step regardless.

cerdfied: in what cases would we want to commit this directory?

ivomurrell: > I believe we shouldn't give the option and perform this step regardless.

The script intentionally doesn't do anything without asking the user first – I believe this makes it easier to trust as well as understand what we're changing.

> in what cases would we want to commit this directory?

None that we support but the worst that will happen is constant churn in the git history with the state changing whenever someone runs Tool Kit and commits it afterwards. Sensitive data is never stored in the state.

joelcarr: As we don't support the use-cases to commit the state data, we shouldn't make it seem like we do by offering the option. Offering the option will mean the question "why would I like to store `.toolkitstate/`?" will be raised. If we are giving the option, we have to have an answer to this question.

ivomurrell: We also don't support using your old CircleCI config with a Tool Kit project, nor having n-gage installed in the same project, or installing both Mocha and Jest in the same project, but the script allows you to opt into all of those options too. Perhaps we should change the script to be more authoritative and not provide prompts for anything we don't support (though I still think politely checking with the user leads to a more positive and empowering experience and that outweighs the costs) but regardless I think that's outside the scope of this PR.

joelcarr (approved):

PR #556, opened 2023-12-18
CPP-1776 Add option to remove ESLint config from gitignore

ivomurrell:
# Description

n-gage would copy a default ESLint config into your project but [only if `.eslintrc.js` was in your `.gitignore`](https://github.com/Financial-Times/n-gage/blob/1edbe16b1f49058b534c0bd117ce5bd0f17d13c2/src/tasks/install.mk#L55). This means that many n-gage projects don't have committed ESLint configuration. The Tool Kit migration script will set up our own recommended ESLint config (pulling in the settings from [eslint-config-next](https://github.com/Financial-Times/eslint-config-next)) but we should also make sure the config isn't also ignored by git.

# Checklist:

### DIFF
--- a/core/create/src/index.ts
+++ b/core/create/src/index.ts
@@ -55,7 +55,7 @@ function clearConfigCache() {
 async function executeMigration(
   deleteConfig: boolean,
   addEslintConfig: boolean,
-  ignoreToolKitState: boolean,
+  fixGitignore: boolean,
   configFile: string
 ): Promise<void> {
   for (const pkg of packagesToInstall) {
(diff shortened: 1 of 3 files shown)


core/create/src/prompts/main.ts
```diff
@@ -114,14 +101,12 @@ export default async ({
...
+        name: 'fixGitignore',
+        type: 'confirm',
         initial: true,
-        message: `Would you like the ${styles.filepath(
-          '.toolkitstate/'
-        )} directory to be added to your ${styles.filepath(
+        message: `Would you like your ${styles.filepath(
           '.gitignore'
-        )}? Tool Kit stores some state files that shouldn't be committed to your repository.`
+        )} to be updated? Tool Kit stores some state files that shouldn't be committed to your repository, whereas ESLint configuration can now be committed.`
```
apaleslimghost, line 109: **nitpick**: minor wording tweak for clarity

```suggestion
        )} to be updated? Tool Kit stores some state files that shouldn't be committed to your repository, whereas ESLint configuration should now be committed.`
```

apaleslimghost (approved):

core/create/src/index.ts
```diff
@@ -82,14 +82,29 @@ async function executeMigration(
       )
     : Promise.resolve()
 
+  const gitignorePath = '.gitignore'
+  const fixGitignoreAsync = async () => {
```
cerdfied, line 86: **nitpick:** maybe we want to include a comment here as to why we are doing this? I was trying to think of a way we could do that via naming but think it's a little more involved than that

ivomurrell: do you mean document what we are doing in this closure (editing the `.gitignore`) or why we're creating an async closure in the first place (so that we can wrap all this logic in one komatsu log)?

cerdfied: I mean the why!

cerdfied (approved):

PR #559, opened 2023-12-21
CPP-1832 add explicit init functions

apaleslimghost:
adds an `Init` class that plugins can point to from `.toolkitrc`. all init classes from all plugins are run on startup, when running `--install` and tasks.

this is intended for things like setting state variables from environment, like in CircleCI. we were previously doing this as side effects in plugin entrypoints, which doesn't work in a post `lazy-plugins`/`even-lazier-plugins` world.

yes ivo i know these classes could just be functions. by making them subclasses of `Base` we get all the version validation stuff for free, and also that's what `importEntryPoint` expects


core/cli/src/install.ts
```diff
@@ -112,11 +113,13 @@ export default async function installHooks(logger: Logger): Promise<ValidConfig>
     }
   }
 
+  await runInit(logger, config)
+
   const errors: Error[] = []
-  // group hooks without an installGroup separately so that their check()
-  // method runs independently
```
apaleslimghost, line 117: this comment got moved away from the thing it was commenting on at some point, so i've just moved it back

plugins/circleci/.toolkitrc.yml
```diff
@@ -1,2 +1,5 @@
 installs:
   CircleCiConfig: './lib/circleci-config'
+
+init:
+  - './lib/env-vars'
```
ivomurrell, line 5: **thought**: maybe the file should include `init` in its name (innit) to make its purpose clearer?

core/cli/src/tasks.ts
```diff
@@ -64,6 +65,7 @@ ${availableHooks}`
   }
 
   await checkInstall(logger, config)
+  await runInit(logger, config)
```
ivomurrell, line 68: **question**: why doesn't `runInit` run before `checkInstall`? initialisation logic would have implicitly run first previously as `checkInstall` loads the plugins

core/cli/src/init.ts
```diff
@@ -0,0 +1,23 @@
...
+import { importEntryPoint } from './plugin/entry-point'
+
+const loadInitEntrypoints = async (logger: Logger, config: ValidConfig): Promise<Validated<Init[]>> => {
+  const initClassResults = reduceValidated(
+    await Promise.all(
+      config.inits.map(async (entryPoint) => {
+        const hookResult = await importEntryPoint(Init, entryPoint)
+        return hookResult.map((initClass) => initClass as InitClass)
+      })
+    )
```
ivomurrell, line 12: **suggestion**: can be simplified by casting the `importEntryPoint` argument instead of mapping over the `Validated` result
```suggestion
    await Promise.all(config.inits.map((entryPoint) => importEntryPoint(Init as InitClass, entryPoint)))
```

ivomurrell (commented):
**question**: would it be possible to only initialise plugins that we're going to lazily load in a given Tool Kit invocation? would that be too complex? just seems a bit weird to _explicitly_ be running the CircleCI initialisation logic for every Tool Kit run even if I'm not worried about a performance impact...

core/cli/src/tasks.ts
apaleslimghost: good point!

apaleslimghost:
> would it be possible to only initialise plugins that we're going to lazily load in a given Tool Kit invocation?

because we're now lazily loading each Task/Hook individually, there's no concept of "plugins that we're going to lazily load". in this branch we do only init plugins that are included somewhere in a config, it's not that we're initialising _everything_ for _everyone_

ivomurrell (approved):

PR #560, opened 2024-01-02
CPP-1833 part one: split validated out of types into its own package

apaleslimghost:


lib/validated/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+We _could_ still use exceptions/rejections for this, along with `Promise.allSettled`. However, with that pattern it would be difficult to follow the data flow of the errors, and easy to accidentally forget to collate some errors, falling into the same issue of throwing one error at a time.
+
+Instead, we model these errors using an explicit return value, `Validated<T>`. This type is a union of `Valid<T>` which represents a correct value, and `Invalid` which represents a problem with a list of `string` reasons. By making this an explicit type, we clearly mark all the functions that can return errors we may want to collate, and make it impossible to use the values without deciding what to do with the errors.
+
+This pattern is modelled after Go's [error handling using multiple return values](https://go.dev/doc/tutorial/handle-errors), Rust's [`Result<T, E>` type](https://doc.rust-lang.org/book/ch09-00-error-handling.html) and Haskell's [`Either a b` type](https://hackage.haskell.org/package/base-4.19.0.0/docs/Data-Either.html).
```
ivomurrell, line 15: **nitpick**: I think the Rust book [subsection](https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html) on error handling with `Result` is a more specific reference to link to

```suggestion
This pattern is modelled after Go's [error handling using multiple return values](https://go.dev/doc/tutorial/handle-errors), Rust's [`Result<T, E>` type](https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html) and Haskell's [`Either a b` type](https://hackage.haskell.org/package/base-4.19.0.0/docs/Data-Either.html).
```

lib/validated/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+```typescript
+validated.map(value => {
+	// do something with value
+	return // something else. replaces the value with this return value if valid, does nothing if invalid
+})
+
+validated.mapError(reasons => {
+	// do something with reasons
+	return // something else. replaces the reasons with this return value if invalid, does nothing if valid
```
ivomurrell, line 43: **nitpick**: took me a while to realise you were including the `return` keyword in the documentation's sentence. I think it would be more readable to move the 'return' into the comment
```suggestion
    /* do something with value */
	// `return` something else. replaces the value with this return value if valid, does nothing if invalid
})

validated.mapError(reasons => {
	/* do something with reasons */
	// `return` something else. replaces the reasons with this return value if invalid, does nothing if valid
```

lib/validated/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Extracting values from a `Validated`
+
+`Validated` has an `.unwrap` method that allows you to extract its value if valid, or throws an error with its reasons if invalid. This should be used as late as possible, when you can't get any further without the value, and need to stop if it's invalid.
+
+```typescript
+validated.unwrap('something was invalid!')
+```
+
+### Grouping an array of `Validated`
```
ivomurrell, line 55: **praise**: these docs all make the functional style of `Validated` much more approachable imo!

ivomurrell (approved):

PR #566, opened 2024-01-05
[MAIN] fix: switch prettier to a peerDep of the plugin and support prettier 3

apaleslimghost:
backport of #565


apaleslimghost:
thwarted again by jest being Normal

PR #570, opened 2024-01-16
WIP move appropriate options from plugin options to task options

apaleslimghost:
also refactoring tasks that currently have to do weird things because all options are plugin wide


lib/schemas/src/tasks.ts
```diff
@@ -1,9 +1,30 @@
...
+  Jest: JestSchema,
+  Mocha: MochaSchema,
+  Node: NodeSchema,
+  Nodemon: NodemonSchema,
+  Pa11y: Pa11ySchema,
+  Prettier: PrettierSchema,
+  TypeScript: TypeScriptSchema,
+  UploadAssetsToS3Schema: UploadAssetsToS3Schema,
+  WebpackSchema: WebpackSchema,
+  SmokeTestSchema: SmokeTestSchema
```
apaleslimghost, line 27: **issue**: @cerdfied the keys here need to exactly match the names of the tasks that the options are for. we should really have some kind of type checking or something for that, not sure how. sorry this wasn't obvious!

```suggestion
  UploadAssetsToS3: UploadAssetsToS3Schema,
  Webpack: WebpackSchema,
  NTest: SmokeTestSchema
```

apaleslimghost: i've rebased and amended your original commits to fix this

plugins/circleci/src/init-env-vars.ts
```diff
@@ -7,7 +7,8 @@ export default class CircleCIEnvVars extends Init {
       branch: process.env.CIRCLE_BRANCH,
       repo: process.env.CIRCLE_PROJECT_REPONAME,
       version: process.env.CIRCLE_SHA1,
-      tag: process.env.CIRCLE_TAG
+      tag: process.env.CIRCLE_TAG,
+      buildNumber: process.env.CIRCLE_BUILD_NUM
```
ivomurrell, line 11: **praise**: this is a much nicer approach to getting the build number for Serverless tasks!

plugins/serverless/src/tasks/teardown.ts
```diff
@@ -56,10 +35,7 @@ export default class ServerlessTeardown extends Task<{ plugin: typeof Serverless
     }
 
     const child = spawn('serverless', args, {
-      env: {
-        ...process.env,
-        ...vaultEnv
-      }
+      env: process.env
```
ivomurrell, line 38: **nitpick**: this is the [default value of `env`](https://nodejs.org/docs/latest-v18.x/api/child_process.html#child_processspawncommand-args-options) so we don't need to pass any options

plugins/typescript/src/tasks/typescript.ts
```diff
@@ -1,21 +1,29 @@
...
-
+export default class TypeScript extends Task<{ task: typeof TypeScriptSchema }> {
   async run(): Promise<void> {
-    // TODO: add monorepo support with --build option
-    const args = [...this.taskArgs]
-    if (this.pluginOptions.configPath) {
-      args.unshift('--project', this.pluginOptions.configPath)
+    const args = []
+
+    // TODO:KB:20240408 refactor this
```
ivomurrell, line 12: **thought**: this seems fine tbh

plugins/typescript/src/tasks/typescript.ts
ivomurrell, line 1: **praise**: I like how we can roll all the tasks into one `Task` class now and use option values to differentiate them instead of inheritance!

plugins/webpack/src/tasks/webpack.ts
```diff
@@ -0,0 +1,34 @@
+import { type WebpackSchema } from '@dotcom-tool-kit/schemas/lib/tasks/webpack'
+import { Task } from '@dotcom-tool-kit/base'
+import { hookFork, waitOnExit } from '@dotcom-tool-kit/logger'
+import { fork } from 'child_process'
+
+const webpackCLIPath = require.resolve('webpack-cli/bin/cli')
+
+export default class Babel extends Task<{ task: typeof WebpackSchema }> {
+  static description = 'build babel'
```
ivomurrell, line 9: **todo**: the copying and the pasting

```suggestion
export default class Webpack extends Task<{ task: typeof WebpackSchema }> {
  static description = 'build webpack'
```

ivomurrell (changes requested):
**praise**: I think this change leaves our plugins in a much nicer place!! One tiny **todo** to fix before this can be merged but otherwise looks great.

core/cli/src/plugin/options.ts
```diff
@@ -15,10 +17,28 @@ export const validatePluginOptions = (logger: Logger, config: ValidPluginsConfig
...
 
     const pluginSchema = PluginSchemas[pluginId]
+
     if (!pluginSchema) {
       logger.silly(`skipping validation of ${pluginId} plugin as no schema can be found`)
+
+      // TODO:KB:20240412 remove legacyPluginOptions in a future major version
+      if (pluginOptions && pluginId in legacyPluginOptions) {
+        const movedToTask = legacyPluginOptions[pluginId]
+        invalidOptions.push([
```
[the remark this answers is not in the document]
ivomurrell: up to you if you want to change the implementation then, I don't feel as strongly if it will mostly improve errors rather than allowing backwards compatibility!

ivomurrell (approved):

plugins/serverless/src/tasks/teardown.ts
apaleslimghost: i think i prefer having it explicit?

PR #572, opened 2024-01-16
fix: look in config.commandTasks for defined commands, not config.hooks 😅

apaleslimghost:
oops

ig that just shows how long it is since i tried running it

### DIFF
--- a/core/cli/src/tasks.ts
+++ b/core/cli/src/tasks.ts
@@ -14,7 +14,6 @@ import { shouldDisableNativeFetch } from './fetch'
 import { runInit } from './init'
 
 type ErrorSummary = {
-  hook: string
   task: string
   error: Error
 }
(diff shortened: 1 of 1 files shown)


core/cli/src/tasks.ts
```diff
@@ -76,33 +75,33 @@ ${availableHooks}`
...
-          hook,
           task: id,
           error: error as Error
         })
       }
     }
 
     if (errors.length > 0) {
-      const error = new ToolKitError(`error running tasks for ${styles.hook(hook)}`)
+      const error = new ToolKitError(`error running tasks for ${styles.hook(command)}`)
```
ivomurrell, line 104: **thought**: we should define a new style for commands!

ivomurrell (approved):

PR #590, opened 2024-01-29
Recommend forcing npx to use the latest version of the migration script

ivomurrell:
# Description

I always include the `@latest` specifier when people ask the easiest way to migrate to Tool Kit these days, but I never got round to updating the documentation so that this wasn't hidden knowledge. Finally pushed by @apaleslimghost posting about this paper cut!

# Checklist:

### DIFF
--- a/docs/migrating-to-tool-kit.md
+++ b/docs/migrating-to-tool-kit.md
@@ -28,13 +28,15 @@ to follow along with.
 The migration tool is published to npm as `@dotcom-tool-kit/create`. Thanks to
 some [syntax sugar
 magic](https://docs.npmjs.com/cli/v7/commands/npm-init#description) provided by
-npm's `npm init` script this means you can just run
+npm's `npm init` script this means you can run
 
 ```shell
-npm init @dotcom-tool-kit
+npm init @dotcom-tool-kit@latest
 ```
 
-within your repository to fetch the tool and start migrating. Answer the
+within your repository to fetch the tool and start migrating. We include the
+`@latest` specifier here to make sure you're using the latest version of the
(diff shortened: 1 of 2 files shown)


docs/migrating-to-tool-kit.md
```diff
@@ -28,13 +28,15 @@ to follow along with.
 The migration tool is published to npm as `@dotcom-tool-kit/create`. Thanks to
 some [syntax sugar
 magic](https://docs.npmjs.com/cli/v7/commands/npm-init#description) provided by
-npm's `npm init` script this means you can just run
```
apaleslimghost, line 31: **praise**: yessss

scripts/create-plugin.js
```diff
@@ -14,7 +14,7 @@ fs.mkdirSync(directory)
 process.chdir(directory)
 
 console.log('📦 initialising package')
-execSync('npm init -y --scope @dotcom-tool-kit')
+execSync('npm init -y --scope @dotcom-tool-kit@init')
```
apaleslimghost, line 17: **chore**: this is unrelated (and will probably break!), this call is just for creating the `package.json` in a new plugin

**thought**: do we still use this script? does it even still work? maybe let's just delete it?

apaleslimghost (approved):

scripts/create-plugin.js
ivomurrell: oops yeah, thrown off by the scope argument! I feel like we only don't use this script because we aren't in the habit of creating new plugins 😅 but I think we should keep it, if not just to have a bit of documentation on all the things you need to look out for when creating a plugin.

PR #596, opened 2024-02-06
CPP-1428 Fix waiting-for-approval job sometimes running for release builds

ivomurrell:
# Description

We've previously seen a few issues where tag filters were being added to the `waiting-for-approval job` incorrectly (we do not to wait for approval on releases.)

This was happening on some older repositories as they had an older version of Tool Kit installed that still used the `postInstall` function, which has since been [superseded by install state sharing](https://github.com/Financial-Times/dotcom-tool-kit/pull/299) and is deprecated. However, we saw this `postInstall` function was being run with newer versions of Tool Kit too, even though it's supposed to be [skipped](https://github.com/Financial-Times/dotcom-tool-kit/blob/1223562780bd18787b01421adb944615be5976a4/core/cli/src/install.ts#L66-L68)
if support for state sharing is detected. `postInstall` will unconditionally add the tag filter, which is no longer desirable for new CircleCI configs.

It turned out that this was only happening if you ran `dotcom-tool-kit --install` on a repository that already had all its configuration up to date (i.e., all the `check` methods on hooks were returning `true`). This was because the code to confirm that the legacy `postInstall` function can be skipped was inside a block of code only executed when `check`'s were returning `false`. I can't _think_ of any reason why we can't move it out of this conditional block and thus fix Tool Kit's installation behaviour when running `--install` twice consecutively. So I've done that.

The tag filters will usually have to be manually removed from projects which have already inadvertently added them, as they will only be automatically deleted when `dotcom-tool-kit` detects the configuration has gone out of sync (but additional lines added to the CircleCI config (such as the tag filter) are not considered to mean the configuration has gone 'out of sync').

# Checklist:

### DIFF
--- a/core/cli/src/install.ts
+++ b/core/cli/src/install.ts
@@ -25,14 +25,10 @@ export default async function installHooks(logger: Logger): Promise<ValidConfig>
   }
 
   const errors: Error[] = []
-  // HACK: achieve backwards compatibility with older versions of the circleci
-  // plugin that required a postinstall function to run instead of the new
-  // commitInstall method. remove in major update of cli.
-  let usesNewCircleCIGroup = false
   // group hooks without an installGroup separately so that their check()
   // method runs independently
   const groups = groupBy(config.hooks, (hook) => hook.installGroup ?? '__' + hook.id)
-  for (const [groupId, group] of Object.entries(groups)) {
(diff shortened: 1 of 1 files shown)


cerdfied (approved):
makes sense - nice! 🚀

PR #598, opened 2024-02-06
Allow Tool Kit option to override Doppler secret when setting Cypress base URL

ivomurrell:
# Description

We should be able to get the Cypress base URL from either Doppler or the Tool Kit config, but the setting in your config should override what's in Doppler as this will allow users to test out a different URL in a separate branch if needed. [Highlighted](https://financialtimes.slack.com/archives/C3TJ6KXEU/p1707159162412489) by @debugwand.

This is a breaking change as people may have been relying on the previous override behaviour (although only cp-content-pipeline seems to be doing so at this time). 

# Checklist:

### DIFF
--- a/plugins/cypress/src/tasks/cypress.ts
+++ b/plugins/cypress/src/tasks/cypress.ts
@@ -15,7 +15,7 @@ export class CypressLocal extends Task<typeof CypressSchema> {
     const vault = new DopplerEnvVars(this.logger, 'dev')
     const vaultEnv = await vault.get()
 
-    const env = { ...process.env, ...cypressEnv, ...vaultEnv }
+    const env = { ...process.env, ...vaultEnv, ...cypressEnv }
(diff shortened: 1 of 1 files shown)


plugins/cypress/src/tasks/cypress.ts
```diff
@@ -15,7 +15,7 @@ export class CypressLocal extends Task<typeof CypressSchema> {
     const vault = new DopplerEnvVars(this.logger, 'dev')
     const vaultEnv = await vault.get()
 
-    const env = { ...process.env, ...cypressEnv, ...vaultEnv }
+    const env = { ...process.env, ...vaultEnv, ...cypressEnv }
```
cerdfied, line 18: **suggestion**: can we rename these vault variables to denote doppler instead whilst we're here?

cerdfied (approved):

PR #611, opened 2024-04-02
docs(serverless): correct the regions option documentation

rowanmanning:
# Description

The `regions` option was renamed but the documentation wasn't updated alongside it. I've also corrected the fact that this option isn't required – it defaults to `['eu-west-1']` in the code.

# Checklist:


joelcarr (approved):
Thank you 🙏

PR #612, opened 2024-04-04
fix: remove check for undefined commands

apaleslimghost:
since there's no concept of "defining" a command like we had with hooks, we can remove this.


apaleslimghost:
hmm maybe i should have done this in vs code

PR #614, opened 2024-04-08
style: apply prettier to everything

apaleslimghost:
i... have our pre commit hooks not been working???


ivomurrell (approved):
not sure if there actually enough files changed to warrant breaking out the `.git-blame-ignore-revs` file but why not lol

PR #615, opened 2024-04-10
Slightly more backwards compatible

apaleslimghost:
this Should<sup>TM</sup> allow most<sup>[citation needed]</sup> apps' toolkitrcs to load without any changes, just by updating Tool Kit

- renames the new .toolkitrc field `hooks` to `options.hooks`
- reënables specifying the old-style `hooks` in .toolkitrc, treating it as equivalent to the new `commands`, printing a warning

### DIFF
--- a/core/cli/src/plugin/merge-commands.ts
+++ b/core/cli/src/plugin/merge-commands.ts
@@ -2,10 +2,24 @@ import type { CommandTask, Plugin } from '@dotcom-tool-kit/plugin'
 import type { ValidPluginsConfig } from '@dotcom-tool-kit/config'
 import { Conflict, isConflict } from '@dotcom-tool-kit/conflict'
 import { isDescendent } from './is-descendent'
+import { Logger } from 'winston'
+import { styles as s } from '@dotcom-tool-kit/logger'
+import path from 'path'
 
-export const mergeCommands = (config: ValidPluginsConfig, plugin: Plugin) => {
+export const mergeCommands = (config: ValidPluginsConfig, plugin: Plugin, logger: Logger) => {
   if (plugin.rcFile) {
-    for (const [id, configCommandTask] of Object.entries(plugin.rcFile.commands)) {
+    let commands = plugin.rcFile.commands
+
+    if (plugin.rcFile.hooks) {
+      commands = plugin.rcFile.hooks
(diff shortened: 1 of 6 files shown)


core/cli/src/plugin/merge-commands.ts
```diff
@@ -2,10 +2,24 @@ import type { CommandTask, Plugin } from '@dotcom-tool-kit/plugin'
...
+import { styles as s } from '@dotcom-tool-kit/logger'
+import path from 'path'
 
-export const mergeCommands = (config: ValidPluginsConfig, plugin: Plugin) => {
+export const mergeCommands = (config: ValidPluginsConfig, plugin: Plugin, logger: Logger) => {
   if (plugin.rcFile) {
-    for (const [id, configCommandTask] of Object.entries(plugin.rcFile.commands)) {
+    let commands = plugin.rcFile.commands
+
+    if (plugin.rcFile.hooks) {
```
ivomurrell, line 13: **nitpick**: could you add a TODO or HACK comment here so that it's easier to remember to delete this in a future major version?

lib/plugin/src/index.ts
```diff
@@ -3,11 +3,12 @@ export type RCFile = {
   installs: { [id: string]: string }
   tasks: { [id: string]: string }
   commands: { [id: string]: string | string[] }
+  hooks?: { [id: string]: string | string[] }
```
ivomurrell, line 6: **question**: I'm a bit confused about this change – what's the difference between `hooks` and `options.hooks`? does this mean we should move `hooks` into the `options` field in our config files? why do some places still use `hooks` whilst others have deleted it in favour of `options.hooks`?

apaleslimghost: | field | `main` | `next` | this branch |
|-|-|-|-|
| `hooks` | assigning tasks to hooks | specifying hook installation options | assigning tasks to commands (with warning) |
| `commands` | ❌  | assigning tasks to commands | assigning tasks to commands | 
| `options.hooks` | ❌ | ❌ | specifying hook installation options |

ivomurrell: aha, makes perfect sense, thank you!

ivomurrell (approved):

PR #616, opened 2024-04-10
Even more backwards compatible (well sort of)

apaleslimghost:
- adds a `version` field to `.toolkitrc`
- refuses to load a (non-app-root) plugin that doesn't match the const for the currently supported version (now 2)

this should provide users with more actionable error messaging if they have outdated plugins, instead of trying to load an old toolkitrc expecting the current structure and which will probably throw some `cannot read property of undefined` error at some point


core/cli/src/plugin.ts
```diff
@@ -41,6 +41,16 @@ export async function loadPlugin(
...
   }
 
+  if (!isAppRoot && plugin.rcFile.version !== CURRENT_RC_FILE_VERSION) {
+    return invalid([
+      `plugin ${s.plugin(id)} has a v${s.code((plugin.rcFile.version ?? 1).toString())} ${s.code(
+        '.toolkitrc.yml'
+      )}, but this version of Tool Kit can only load v${s.code(
+        CURRENT_RC_FILE_VERSION.toString()
+      )}. please update this plugin.`
+    ])
```
joelcarr, line 51: **thought:** Some assisting documentation to add as part of the major release to explain why this was done.

joelcarr (approved):

core/cli/src/plugin.ts
apaleslimghost: yep. documentation (including a migration guide) will be coming later.

ivomurrell (approved):
good idea!

PR #617, opened 2024-04-10
Backwards compatible .toolkitrc options field

apaleslimghost:
old `.toolkitrc`s contain `options` with just plugin options. these are now nested under `options.plugins`. if an `options` is non-empty, but doesn't contain any of the new top-level fields, assume it's old-style, and nest it.

**nb** this probably won't actually work in most cases, as the options are all moving around in #570. but we'll at least get semi-useful schema errors. maybe?

### DIFF
--- a/core/cli/src/rc-file.ts
+++ b/core/cli/src/rc-file.ts
@@ -18,9 +18,11 @@ let rootConfig: string | undefined
 type RawRCFile = {
   [key in Exclude<keyof RCFile, 'options'>]?: RCFile[key] | null
 } & {
-  options: {
-    [key in keyof RCFile['options']]?: RCFile['options'][key] | null
-  }
+  options:
+    | {
+        [key in keyof RCFile['options']]?: RCFile['options'][key] | null
+      }
+    | null
 }
 
 export async function loadToolKitRC(logger: Logger, root: string, isAppRoot: boolean): Promise<RCFile> {
@@ -44,6 +46,27 @@ export async function loadToolKitRC(logger: Logger, root: string, isAppRoot: boo
   }
 
   const config: RawRCFile = result.config
+
+  // if a toolkitrc contains a non-empty options field, but not options.{plugins,tasks,hooks},
+  // assume it's an old-style, plugins-only options field.
+  if (
+    config.options &&
+    Object.keys(config.options).length > 0 &&
+    !(config.options.plugins || config.options.tasks || config.options.hooks)
+  ) {
+    config.options = {
+      plugins: config.options as { [id: string]: Record<string, unknown> }
+    }
+
+    logger.warn(
+      `plugin at ${s.filepath(path.dirname(root))} has an ${s.code(
+        'options'
+      )} field that only contains plugin options. these options should be moved to ${s.code(
+        'options.plugins'
+      )}.`
+    )
+  }
+
   return {
     plugins: config.plugins ?? [],
     installs: config.installs ?? {},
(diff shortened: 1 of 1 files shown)


core/cli/src/rc-file.ts
```diff
@@ -44,6 +46,27 @@ export async function loadToolKitRC(logger: Logger, root: string, isAppRoot: boo
   }
 
   const config: RawRCFile = result.config
+
+  // if a toolkitrc contains a non-empty options field, but not options.{plugins,tasks,hooks},
```
ivomurrell, line 50: **nitpick**: same as [#615](https://github.com/Financial-Times/dotcom-tool-kit/pull/615#discussion_r1559157732), could you add a TODO or HACK comment here so that it's easier to remember to delete this in a future major version?

core/cli/src/rc-file.ts
```diff
@@ -44,6 +46,27 @@ export async function loadToolKitRC(logger: Logger, root: string, isAppRoot: boo
...
+  // assume it's an old-style, plugins-only options field.
+  if (
+    config.options &&
+    Object.keys(config.options).length > 0 &&
+    !(config.options.plugins || config.options.tasks || config.options.hooks)
+  ) {
+    config.options = {
+      plugins: config.options as { [id: string]: Record<string, unknown> }
```
ivomurrell, line 58: **issue**: will this work for options that are now task options instead of plugin options?

apaleslimghost: it will not. there's not going to be an easy way to automatically handle those in general. this will let those options fall through to the plugin option validator, which it looks like will just skip plugins that don't have a schema. what i could do is in #570 keep the schemas i deleted but change them all to `z.never()`, which should fail validation with a hopefully useful error message?

ivomurrell: yeah, that sounds like a good compromise

core/cli/src/rc-file.ts
```diff
@@ -44,6 +46,28 @@ export async function loadToolKitRC(logger: Logger, root: string, isAppRoot: boo
...
+    }
+
+    logger.warn(
+      `plugin at ${s.filepath(path.dirname(root))} has an ${s.code(
+        'options'
+      )} field that only contains plugin options. these options should be moved to ${s.code(
+        'options.plugins'
+      )}.`
```
ivomurrell, line 67: **suggestion**: clarify that some options will have to be moved to `options.tasks` instead. I think  it will be surprising if a user still gets errors after following the suggestion here. I've tried my best to make it less surprising though it's hard to write something clear for this 😅

```suggestion
      )} field that only contains plugin options. these options should be moved to ${s.code(
        'options.plugins'
      )} (or ${s.code('options.tasks')} if you see subsequent errors.)`
```

ivomurrell (approved):

core/cli/src/rc-file.ts
apaleslimghost: done in https://github.com/Financial-Times/dotcom-tool-kit/pull/570/commits/cb98a854a582fe6cb76a5c57156ff5101afcb1bf

core/cli/src/rc-file.ts
apaleslimghost: actually, as well as this warning, with https://github.com/Financial-Times/dotcom-tool-kit/commit/cb98a854a582fe6cb76a5c57156ff5101afcb1bf they'll also have got a bunch of schema errors already. i think it's fine?

ivomurrell: will they see those schema errors before they move to `options.plugins`?

apaleslimghost: yes

PR #618, opened 2024-04-11
CPP-1834 command task options

apaleslimghost:
allows loading task options from command task assignments, like this:

```yml
commands:
   'test:review':
     - Jest:
          configFile: 'jest.e2e.config.js'
```

these are merged with options from `options.tasks` to override them.

there's probably some work to do around the ergonomics of the schema error messages but whatever


core/cli/src/help.ts
```diff
@@ -50,7 +50,7 @@ ${(Hook.constructor as any).description ? (Hook.constructor as any).description
 ${
   tasks && tasks.tasks.length
     ? `runs ${tasks.tasks.length > 1 ? 'these tasks' : 'this task'}:
-${tasks.tasks.map((task) => `- ${styles.task(task)}`).join('\n')}`
+${tasks.tasks.map((task) => `- ${styles.task(task.task)}`).join('\n')}`
```
cerdfied, line 53: **suggestion**: can we maybe refactor this so its not `tasks.tasks`?

apaleslimghost: hmm wow yeah this code is horrible to read 😂 i'll see what i can do! although tbh, the code for these messages isn't super important, i'm not particularly fussed if they're a bit clunky

ivomurrell (approved):

core/cli/src/help.ts
apaleslimghost: actually we need to rework `--help` anyway so i'll leave this for now

PR #619, opened 2024-04-11
CPP-1747 CPP-1748 CPP-1822 Define CircleCI configuration declaratively in .toolkitrc.yml

ivomurrell:
# Description

Following on from similar changes to the `PackageJson` hook, the `CircleCI` hook is now configured via Tool Kit options in `.toolkitrc.yml` files instead of modifying shared state in JavaScript functions. This creates a declarative approach instead of the old functional approach, which should make configuration much easier to reason with and less abstract, as well as more flexible, allowing for different kinds of workflows.

With this initial PR I've focused on getting the configuration generated for [next-static](https://github.com/Financial-Times/next-static) to be functionally equivalent to what was generated previously. Subsequent work will be done to make the config generated better than what was previously generated, and to support a more diverse range of project types.

Off the top of my head, the outstanding things that need to be done in further PRs are

1. Define CircleCI jobs directly in a project's config rather than offloading to our orb. This will reduce a level of abstraction and allow users to directly edit the steps in the jobs they use.
2. Add support for conditional options in Tool Kit configs based on Tool Kit options using YAML custom tags. This will allow the Serverless and Cypress code paths to be defined declaratively.
3. Improve conflict resolution to catch more conflict issues.
4. Simplify generated config when a project doesn't need to make use of matrices of Node versions.

_Please_ point out if I've left any gaps in the remaining work as I tend to develop an [SEP field](https://en.wikipedia.org/wiki/Somebody_else's_problem#Douglas_Adams'_SEP_field) once I work on the same code for too long.

# Checklist:


plugins/circleci/test/circleci-config.test.ts
apaleslimghost, line 1: **praise**: nice comprehensive tests! think they cover everything i'd expect

plugins/circleci/src/circleci-config.ts
```diff
@@ -293,33 +200,150 @@ const getInitialState = (): CircleCIState => {
...
+  CircleCiOptions,
+  'custom'
+>)[]
+
 const isAutomatedConfig = (config: string): boolean => config.startsWith(automatedComment)
 
 const isNotToolKitConfig = (config: string): boolean => !config.includes('tool-kit')
 
-const getJobName = (job: Job): string => (typeof job === 'string' ? job : Object.keys(job)[0])
+const isObject = (val: unknown): val is Record<string, unknown> => isPlainObject(val)
```
apaleslimghost, line 212: **thought**: we keep writing this predicate, presumably there's a type predicates library somewhere we can use?

plugins/circleci/src/circleci-config.ts
apaleslimghost, line 1: **suggestion**: as we discussed on slack this file is a bit hard to follow and could do with restructuring/splitting up for readability

apaleslimghost (approved):
been testing this and it works as i'd expect, i don't fully follow the conflict resolution code but i'm happy that the test cases for it accurately explain the intended logic so does it matter really? nice work!

plugins/circleci/test/circleci-config.test.ts
```diff
@@ -18,67 +25,66 @@ jest.mock('fs', () => {
...
-    setOptions('@dotcom-tool-kit/circleci', { nodeVersion: '16.14-browsers' })
+    setOptions('@dotcom-tool-kit/circleci', { cimgNodeVersions: ['16.14-browsers'] })
   })
 
   afterEach(() => {
     process.chdir(originalDir)
   })
 
-  describe('check', () => {
+  describe.skip('check', () => {
```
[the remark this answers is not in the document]
ivomurrell: oh yeah I was skipping these tests because there's now no longer a nice way to persuade the hook to discard the base configuration that isn't relevant for the tests (like defining the `checkout` job, pulling in the orbs, etc). Jest mocks are a mess and don't work for this case. but maybe it would be useful to disable to the base config for some projects so might make sense to be an option we can set...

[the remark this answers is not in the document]
apaleslimghost: ah i see, yeah that's a pain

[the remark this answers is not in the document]
apaleslimghost: could you include a comment explaining why the test is skipped in that case please

plugins/circleci/src/circleci-config.ts
ivomurrell: 🦄 https://github.com/sindresorhus/is#objectvalue

plugins/circleci/test/circleci-config.test.ts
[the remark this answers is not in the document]
ivomurrell: I think I've convinced myself that it's more a sign that we should allow the base config to be disabled. I'll make a separate commit to add that

plugins/circleci/src/circleci-config.ts
apaleslimghost: Herr Sorhus, you've done it again!

PR #625, opened 2024-04-16
CPP-1823 Get list of files to be hashed from the hooks

apaleslimghost:


core/cli/src/config/hash.ts
```diff
@@ -17,20 +18,26 @@ export async function fileHash(path: string): Promise<string> {
...
-export async function hasConfigChanged(logger: Logger): Promise<boolean> {
+export async function hasConfigChanged(logger: Logger, config: ValidConfig): Promise<boolean> {
   const hashes = readState('install')
+
   if (!hashes) {
     return true
   }
-  for (const [path, prevHash] of Object.entries(hashes)) {
+
+  for (const path of ['.toolkitrc.yml', ...config.hookManagedFiles]) {
```
ivomurrell, line 37: **suggestion(non-blocking)**: think we could also consider the config to have changed if `Object.keys(hashes) !== ['.toolkitrc.yml', ...config.hookManagedFiles]` to catch cases where a hook has been removed. though idk if that would ever require a reinstall?

plugins/circleci/.toolkitrc.yml
```diff
@@ -1,5 +1,6 @@
 installs:
-  CircleCi: './lib/circleci-config'
+  CircleCi:
+    entryPoint: './lib/circleci-config'
```
ivomurrell, line 3: **suggestion(non-blocking)**: should we add the managed files as part of this PR?

```suggestion
    entryPoint: './lib/circleci-config'
    managedFiles:
      - '.circleci/config.yml'
```

plugins/package-json-hook/.toolkitrc.yml
```diff
@@ -1,4 +1,5 @@
 installs:
-  PackageJson: './lib/package-json-helper'
+  PackageJson:
+    entryPoint: './lib/package-json-helper'
```
ivomurrell, line 3: **suggestion(non-blocking)**: should we add the managed files as part of this PR?

```suggestion
    entryPoint: './lib/package-json-helper'
    managedFiles:
      - 'package.json'
```

ivomurrell (approved):
nice!

plugins/circleci/.toolkitrc.yml
apaleslimghost: oops, meant to do that but forgot!

core/cli/src/config/hash.ts
apaleslimghost: it's impossible for a _removed_ hook to cause a difference with install so i don't think we need this

PR #626, opened 2024-04-19
CPP-2041 CPP-2030 Add support for tags in Tool Kit config that resolve to values of options

ivomurrell:
# Description

These two custom YAML tags allow us to define all the different combinations of Serverless and Cypress configuration needed to be added to CircleCI config. These were previously defined in JavaScript functions that were hard to compose but now they can be specified in a declarative, readable syntax along with the rest of the config.

There are two tags available currently: `!toolkit/option` will substitute the specified option as a string into your config, whilst `!toolkit/defined` will conditionally include its value in your config if the specified option is defined. You can see how they can be used in the test cases and in `circleci-deploy`'s updated config (scroll down all the way to the end of the PR for that!)

# Checklist:

### DIFF
--- a/core/cli/src/plugin/options.ts
+++ b/core/cli/src/plugin/options.ts
@@ -1,10 +1,15 @@
 import { ValidPluginsConfig } from '@dotcom-tool-kit/config'
-import { InvalidOption } from '../messages'
-import { type PluginOptions, PluginSchemas, legacyPluginOptions } from '@dotcom-tool-kit/schemas'
 import { isConflict } from '@dotcom-tool-kit/conflict'
+import { ToolKitError } from '@dotcom-tool-kit/error'
+import { OptionsForPlugin, RCFile, type Plugin } from '@dotcom-tool-kit/plugin'
+import { type PluginOptions, PluginSchemas, legacyPluginOptions } from '@dotcom-tool-kit/schemas'
+
 import type { Logger } from 'winston'
 import { ZodError, ZodIssueCode } from 'zod'
 import { styles } from '@dotcom-tool-kit/logger'
+
+import { toolKitIfDefinedIdent, toolKitOptionIdent } from '../rc-file'
+import { InvalidOption } from '../messages'
 
 export const validatePluginOptions = (logger: Logger, config: ValidPluginsConfig): InvalidOption[] => {
   const invalidOptions: InvalidOption[] = []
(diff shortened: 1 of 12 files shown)


core/create/src/index.ts
```diff
@@ -44,16 +43,6 @@ function getEslintConfigContent(): string {
...
 
-function clearConfigCache() {
-  // we need to import explorer from the app itself instead of npx as this is
-  // the object used by installHooks()
-  return (
-    importCwd('dotcom-tool-kit/lib/rc-file') as {
-      explorer: ReturnType<typeof cosmiconfig>
-    }
-  ).explorer.clearSearchCache()
-}
```
apaleslimghost, line 55: **praise**: thank fuck

core/cli/src/rc-file.ts
```diff
@@ -25,27 +26,54 @@ type RawRCFile = {
...
+// minimally define the two custom tags so that yaml will parse them without
+// warning but will never be use them when stringifying
+const toolKitOption = {
+  identify: () => false,
+  tag: '!toolkit/option',
+  resolve: (value) => value
+} satisfies YAML.ScalarTag
+const toolKitDefined = {
+  identify: () => false,
+  tag: '!toolkit/defined',
```
apaleslimghost, line 43: **suggestion(bikeshedding)**: i think `!toolkit/if-defined` (or maybe `if-option-defined`? bit verbose tho) would be clearer in intent and more readable

core/cli/src/rc-file.ts
```diff
@@ -25,27 +26,54 @@ type RawRCFile = {
...
   }
 
-  const config: RawRCFile = result.config
+  const configDoc = YAML.parseDocument(rawConfig, { customTags: [toolKitOption, toolKitDefined] })
+  // go back and search for the parsed tags and replace them with our
+  // indentifiers so we can resolve them once we've loaded plugin options
+  YAML.visit(configDoc, {
+    Scalar(_, node) {
+      if (node.tag === '!toolkit/option') {
+        return configDoc.createNode({ [toolKitOptionIdent]: node.value })
```
apaleslimghost, line 65: **question**: why is this done in a `YAML.visit` and not in the `resolve` method of the tags? i did a quick test of that and it preserves symbols as well:

Screenshot 2024-04-22 at 15 07 23

core/cli/src/plugin/options.ts
```diff
@@ -59,3 +64,72 @@ export const validatePluginOptions = (logger: Logger, config: ValidPluginsConfig
...
+  // recursively walk through the parsed config, searching for the tag
+  // identifiers we've inserted during parsing, and substitute them for
+  // resolved option values
+  const deeplySubstitute = (node: unknown, path: (string | number)[]): unknown => {
+    if (Array.isArray(node)) {
+      return node.map((item, i) => deeplySubstitute(item, [...path, i]))
+    } else if (node && typeof node === 'object') {
+      const entries = Object.entries(node)
+      // !toolkit/option will be marked by a single key within an object
+      if (entries[0]?.[0] === toolKitOptionIdent) {
```
apaleslimghost, line 94: **suggestion**: i think having an example here in a comment of what the YAML and the resulting shape of the object in the config look like would help for people trying to understand this function

core/cli/src/plugin/options.ts
```diff
@@ -59,3 +64,72 @@ export const validatePluginOptions = (logger: Logger, config: ValidPluginsConfig
...
+    return (config.pluginOptions[pluginName] as OptionsForPlugin)?.options[optionName]
+  }
+
+  // throw an error if there are tags in plugin option fields to avoid circular
+  // references
+  const validateTagPath = (path: (string | number)[]) => {
+    if (path[0] === 'options' && path[1] === 'plugins') {
+      const error = new ToolKitError('cannot reference plugin options when specifying options')
+      error.details = `YAML tag referencing options used at path '${path.join('.')}'`
+      throw error
```
apaleslimghost, line 81: **suggestion**: use `Validated` for this to avoid whackamole if they've done it multiple times?

core/cli/src/rc-file.ts
apaleslimghost: **suggestion**: actually, instead of using objects with a specific key here, we could use classes with a property for the option key and the body of the config to substitute. that would also avoid having to concatenate the option key for `toolkit/defined`. doing this would make `deeplySubstitute` a lot cleaner, you could do things like this instead of having to sniff around the object entries:

```js
if(node instanceof OptionsPlaceholder) {
  return resolveOptionPath(node.key)
}
```

apaleslimghost (approved):
this is great! i have a few suggestions, including one fairly big one that will help with clarity but we could do as a refactor later on

core/cli/src/rc-file.ts
ivomurrell: this might work for `!toolkit/option`, but I don't think it will for `!toolkit/is-defined` 😭 `is-defined` is currently tagging the key of a key-value pair within an object (or a 'map' in YAML nomenclature) and whatever is returned from its `resolve` callback is placed back in the key's position and subsequently stringified (along with a warning) when converted to a JS object. so we can't stuff an object in like we can for `!toolkit/option` and even a symbol is stringified, making these identifiers the least-worst solution. there may be a work around that limitation by finding another part of the YAML schema to tag with `!toolkit/is-defined`, but during my investigation i couldn't find anywhere else that would result in a readable, declarative syntax like we have now.

core/cli/src/plugin/options.ts
```diff
@@ -74,48 +74,65 @@ export const substituteOptionTags = (plugin: Plugin, config: ValidPluginsConfig)
 
   // throw an error if there are tags in plugin option fields to avoid circular
   // references
-  const validateTagPath = (path: (string | number)[]) => {
+  const validateTagPath = (path: (string | number)[]): Validated<void> => {
```
apaleslimghost, line 77: **suggestion**: to avoid the casting to `Validated<any>`, this could just return `boolean`. `substituted.push(invalid(['whatever']))` will infer the type correctly

core/cli/src/plugin/options.ts
```diff
@@ -74,48 +74,65 @@ export const substituteOptionTags = (plugin: Plugin, config: ValidPluginsConfig)
...
-              Object.assign(substituted, deeplySubstitute(value, path))
-            } // don't include the node if !optionValue
+            // keep walking the path if we've found an error here so we can
+            // gather even more errors to show the user. else skip traversal if
+            // we aren't going to include the node
+            if (optionValue || !validatedKey.valid) {
+              const subbedValues = deeplySubstitute(value, path)
+              if (subbedValues.valid) {
+                substituted.push(...Object.entries(subbedValues.value as object).map((v) => valid(v)))
+              } else {
```
apaleslimghost, line 121: **suggestion**: i _think_ this could use `validated.flatMap`?

apaleslimghost (approved):
couple more suggestions, thanks for the refactor!

core/cli/src/plugin/options.ts
ivomurrell: it could use `validated.map` (not `validated.flatMap` because we want `Validated<T>[]` not `Validated<T[]>` here) but I think it makes it less readable in this case

PR #627, opened 2024-04-22
CPP-2040 update --help for next

apaleslimghost:
screenshots of various scenarios:

<details><summary>kitchen sink</summary>

Screenshot 2024-04-22 at 14 37 36

</details>

<details><summary>commands but no hooks</summary>

Screenshot 2024-04-22 at 14 37 54

</details>

<details><summary>hooks but no commands</summary>

Screenshot 2024-04-22 at 14 48 56

</details>

<details><summary>asking for help for specific commands</summary>

n.b. in this example and the following two, there is still a hook installed; i decided if you're asking about particular command, information about hooks is irrelevant

Screenshot 2024-04-22 at 14 37 24

</details>

<details><summary>asking for help for a command that doesn't exist</summary>

Screenshot 2024-04-22 at 14 37 13

</details>

<details><summary>asking for help for both commands that exist and commands that don't</summary>

Screenshot 2024-04-22 at 14 37 00

</details>

commits in this one are a bit of a mess tbh. i'd mostly just ignore the commits and the diff and review it on the basis of the new code.


lib/logger/src/styles.ts
```diff
@@ -16,9 +21,22 @@ export const styles = {
   dim: colours.grey,
   title: colours.bold.underline,
   taskHeader: colours.bgWhite.black,
-  errorHighlight: colours.red,
-  error: (string: string): string => `${styles.errorHighlight.bold('‼︎')} ${styles.title(string)}`,
-  warningHighlight: colours.yellow,
-  warning: (string: string): string => styles.warningHighlight.bold('⚠︎') + ' ' + string,
-  ruler: (): string => styles.dim('─'.repeat(process.stdout.columns / 2))
+  errorHighlight: colours.bgRed.bold,
+  error: (string: string): string => `${styles.errorHighlight(' × ️')} ${styles.title(string)}`,
```
ivomurrell, line 25: **todo**: there is a [cursed character](https://stackoverflow.com/questions/38100329/what-does-u-ufe0f-in-an-emoji-mean-is-it-the-same-if-i-delete-it) hidden after the multiplication sign which seems to confuse most text renderers (you can see the monospacing is broken in the diff on GitHub though it absolutely destroys my combo of tmux + nvim) probably meant to denote that the multiplication should be rendered as emoji but something seems to have gone wrong and the two code points have been split by a space.

you could either remove the code point
```suggestion
  error: (string: string): string => `${styles.errorHighlight(' × ')} ${styles.title(string)}`,
```

or fix the split between the two code points (though i'm not sure if the selector will have any effect on the multiplication character, i don't see it listed in this [table](http://unicode.org/emoji/charts/emoji-variants.html))
```suggestion
  error: (string: string): string => `${styles.errorHighlight(' ×️ ')} ${styles.title(string)}`,
```

or maybe use a different emoji 😄
```suggestion
  error: (string: string): string => `${styles.errorHighlight(' ✖️ ')} ${styles.title(string)}`,
```

lib/logger/src/styles.ts
```diff
@@ -16,9 +21,22 @@ export const styles = {
...
+  helpHighlight: colours.bgGreen.bold,
+  help: (string: string): string => styles.helpHighlight(' ? ') + ' ' + string,
+  ruler: (): string => styles.dim('─'.repeat(process.stdout.columns / 2)),
+  box: (string: string, options: Partial<BoxenOptions>) =>
+    boxen(string, {
+      borderStyle: 'round',
+      padding: { left: 1, right: 1 },
+      ...options
+    }),
+  groupHeader: (string: string) => ` ╭─${'─'.repeat(stripAnsi(string).length)}─╮
```
ivomurrell, line 39: **nitpick**: this length calculation logic [won't be able to handle emojis or non-english characters](https://tonsky.me/blog/unicode/), and we should technically be using a library like [`string-width`](https://github.com/sindresorhus/string-width). buuuuuuut we're probably only going to be using ASCII right?

ivomurrell, line 39: **issue**: i get misaligned headers if my terminal window is too small (i was testing the code in a tmux pane 80 characters wide.) i suspect this is a bug in `boxen` as i can't see why else the space at the beginning of the string would be trimmed.

Screenshot 2024-04-22 at 16 29 17

core/cli/src/help.ts
```diff
@@ -1,13 +1,88 @@
...
+    ${s.URL('https://github.com/financial-times/dotcom-tool-kit')}
+  `,
+  { headerText: `🧰 ${s.title(`welcome to ${s.app('Tool Kit')}!`)}` }
+)
+
+const formatTask = ({ task, options }: OptionsForTask) => $t`
+  ${s.task(task)}${
+  Object.keys(options).length > 0
+    ? ` ${s.dim('with options:')}
+    ${YAML.stringify(options).trim()}`
```
ivomurrell, line 22: **praise**: lol this is a nice way to produce a familiar formatting structure for the user

core/cli/src/help.ts
```diff
@@ -1,13 +1,88 @@
...
+    : ''
+}`
+
+const formatCommandTasks = (config: ValidConfig, commands: string[]) =>
+  s.box(
+    $t`
+  ${s.help(
+    `${s.command('commands')} run Tool Kit tasks with ${s.code(
+      'npx dotcom-tool-kit $command'
+    )}, or via configuration installed by hooks in your repository.`
```
ivomurrell, line 32: **praise**: i think a lot of people will be helped by putting Tool Kit into context like this

core/cli/src/help.ts
```diff
@@ -1,13 +1,88 @@
...
+  )}
+  ${commands
+    .map((command) =>
+      config.commandTasks[command] ? formatCommandTask(command, config.commandTasks[command]) + '\n' : ''
+    )
+    .join('')}
```
ivomurrell, line 38: **nitpick**: using a filter map would be more readable and less repetitive imo

```suggestion
    .filter((command) => config.commandTasks[command])
    .map((command) => formatCommandTask(command, config.commandTasks[command]))
    .join('\n')}

```

ivomurrell (commented):
this is great! there are a few fun text encoding issues to resolve but i think the resulting help text is very friendly and readable!

lib/logger/src/styles.ts
```diff
@@ -16,9 +21,22 @@ export const styles = {
...
-  warningHighlight: colours.yellow,
-  warning: (string: string): string => styles.warningHighlight.bold('⚠︎') + ' ' + string,
-  ruler: (): string => styles.dim('─'.repeat(process.stdout.columns / 2))
+  errorHighlight: colours.bgRed.bold,
+  error: (string: string): string => `${styles.errorHighlight(' × ️')} ${styles.title(string)}`,
+  warningHighlight: colours.bgYellow.bold,
+  warning: (string: string): string => styles.warningHighlight(' ! ') + ' ' + string,
+  infoHighlight: colours.bgBlueBright.bold,
+  info: (string: string): string => styles.infoHighlight(' i ') + ' ' + string,
+  helpHighlight: colours.bgGreen.bold,
```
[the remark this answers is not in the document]
apaleslimghost: intriguing! pretty sure it is something weird with VS Code. and yeah my font doesn't have bold, and VS Code doesn't do faux bold

lib/logger/src/styles.ts
apaleslimghost: gahhhh. that's probably because i was using `‼︎` at one point which needs the variant selector to not render as 
‼️. `×` isn't a character with variants.

lib/logger/src/styles.ts
apaleslimghost: it looks like it's trimming all the leading whitespace! even from the `?`/`i`. the version of `boxen` i'm using is a fork of an older version of Sindre's library that still seems vaguely maintained. i don't know if the current Sindre version would work better but it's ESM only 🙃

apaleslimghost: ah. let me guess, `string-width` is ESM-only? i'm not sure how bothered i can be about this tbh

ivomurrell: yep, probably not worth it

ivomurrell (approved):
good stuff! rip you having another package under your belt

PR #630, opened 2024-04-24
Monorepo plugin

apaleslimghost:
- just exports a task. if you want to run commands in your workspace packages, run that task at the toplevel, and it will run the same command in the workspace

  <details><summary>example</summary>
  
  #### `.toolkitrc.yml`
  
  ```yml
  commands:
    test:local: WorkspaceCommand
  ```
  
  #### `packages/something/.toolkitrc.yml`
  
  ```yml
  commands:
    test:local: Cypress # only needed in this package
  ```

  `npx dotcom-tool-kit test:local` in the root runs `test:local` in `packages/something`, which runs `Cypress`
  
  </details>

- ~~refactors some stuff in the core to expose some internals needed for the monorepo task~~ → #709 
- collects workspace package command errors into an `AggregateError` to make use of the new(ish) `AggregateError` handling #716


plugins/eslint/src/tasks/eslint.ts
```diff
@@ -1,13 +1,13 @@
...
-import { Task } from '@dotcom-tool-kit/base'
+import { Task, TaskRunContext } from '@dotcom-tool-kit/base'
 import { ESLintSchema } from '@dotcom-tool-kit/schemas/lib/tasks/eslint'
 import { ESLint } from 'eslint'
 
 export default class Eslint extends Task<{ task: typeof ESLintSchema }> {
   static description = ''
 
-  async run(files?: string[]): Promise<void> {
+  async run({ files }: TaskRunContext): Promise<void> {
```
apaleslimghost, line 10: **question(for @ivomurrell)**: typescript doesn't infer the type of this argument, do you know why?

ivomurrell: > It’s important to understand that an `implements` clause is only a check that the class can be treated as the interface type. It doesn’t change the type of the class or its methods _at all_.

https://www.typescriptlang.org/docs/handbook/2/classes.html#cautions

Guess you can imagine that the check that a subclasses properly implements a superclass comes after the subclass's types have been resolved.

ivomurrell (commented):
i'm very happy that this is so simple! though the proof is in the cp-content-pipeline pudding...

plugins/monorepo/src/load-workspace-configs.ts
```diff
@@ -0,0 +1,58 @@
+import { Init, InitContext } from '@dotcom-tool-kit/base'
+import path from 'path'
+import fs from 'fs/promises'
+import { ValidConfig } from '@dotcom-tool-kit/config'
```
ivomurrell, line 4: **todo**: this and `@dotcom-tool-kit/base` in `workspace-command` aren't included in the `package.json`. the `tsconfig.json` doesn't match up either

plugins/monorepo/src/tasks/workspace-command.ts
```diff
@@ -0,0 +1,44 @@
...
+import {minimatch} from 'minimatch'
+import pluralize from 'pluralize'
+import { styles } from '@dotcom-tool-kit/logger'
+import LoadWorkspaceConfigs from '../load-workspace-configs'
+
+
+export default class WorkspaceCommand extends Task<{ task: typeof WorkspaceCommandSchema }> {
+  async run({ command, files }: TaskRunContext) {
+    const configsWithCommand = LoadWorkspaceConfigs.configs.filter(
+      ({config}) => command in config.commandTasks
```
ivomurrell, line 13: **question**: shouldn't this be checking for `this.options.command` if that's defined? likewise for other uses of `command` in this file

ivomurrell (commented):
looks great!!

PR #637, opened 2024-04-29
CPP-1748 Define hook options for all packages previously using PackageJsonHelper hook

ivomurrell:
# Description

Along with adding the hook options so that plugins that use `package-json-hook`'s `PackageJson` hook will now install their commands into the user's `package.json`, I've also fixed any bugs I found with the Tool Kit code that was preventing the installation from running successfully. I used `next-url-mgmt-service` (with the `lint-staged-npm` plugin added) as a test case. 

# Checklist:

### DIFF
--- a/core/cli/bin/run
+++ b/core/cli/bin/run
@@ -1,5 +1,9 @@
 #!/usr/bin/env node
 
+/* eslint-disable @typescript-eslint/no-var-requires --
+ * this is raw CJS so the lint doesn't apply
+ */
+
 const argv = require('minimist')(process.argv.slice(2), {
   boolean: ['help', 'install', 'listPlugins'],
   '--': true
--- a/core/cli/src/plugin/options.ts
+++ b/core/cli/src/plugin/options.ts
@@ -88,48 +88,74 @@ export const substituteOptionTags = (plugin: Plugin, config: ValidPluginsConfig)
       return reduceValidated(node.map((item, i) => deeplySubstitute(item, [...path, i])))
     } else if (node && typeof node === 'object') {
       const entries = Object.entries(node)
-      // !toolkit/option will be marked as a single entry within an object
-      // e.g., { foo: { '__toolkit/option__': 'foo.bar' } }
-      if (entries[0]?.[0] === toolKitOptionIdent) {
-        const validationError = validateTagPath(path)
-        if (validationError) {
-          return invalid([validationError])
-        } else {
-          const optionPath = entries[0][1] as string
-          return valid(resolveOptionPath(optionPath))
-        }
-      } else {
-        const substituted: Validated<[string, unknown]>[] = []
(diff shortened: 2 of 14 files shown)


core/cli/bin/run
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env node
 
+/* eslint-disable @typescript-eslint/no-var-requires --
+ * this is raw CJS so the lint doesn't apply
+ */
```
apaleslimghost, line 5: **thought**: we're not even linting it anyway because it doesn't have a file extension 😅

core/cli/src/plugin/reduce-installations.ts
```diff
@@ -6,6 +6,11 @@ import { HookSchemas, HookOptions } from '@dotcom-tool-kit/schemas'
 import { Conflict, isConflict } from '@dotcom-tool-kit/conflict'
 import { groupBy } from 'lodash'
 
+// TODO:20240429:IM check if we should be passing this as an argument instead
+// of a global – are we ever reducing installations twice in one Tool Kit
+// invocation?
+const reducedInstallations: Set<string> = new Set()
```
apaleslimghost, line 12: **suggestion**: i was thinking this would be in the `config` alongside the `resolutionTrackers` instead of being _global_ global

**nitpick**: `reducedInstallations` isn't quite an accurate name, it's not the installations we're tracking it's plugins that installations come from. maybe `reducedInstallationPlugins`?

plugins/heroku/.toolkitrc.yml
```diff
@@ -11,4 +11,13 @@ tasks:
...
   'cleanup:remote': NpmPrune
 
+options:
+  hooks:
+    - PackageJson:
+        scripts:
+          heroku-postbuild:
+            - 'build:remote'
+            - 'release:remote'
+            - 'cleanup:remote'
```
apaleslimghost, line 21: **suggestion**: would be good to document these in the plugins' readmes

**thought**: maybe we could autogenerate that documentation 👀

core/cli/src/plugin/options.ts
apaleslimghost, line 1: image

apaleslimghost (approved):

core/cli/bin/run
ivomurrell: ah, well my text editor is linting it! (it inferred the filetype from the shebang)

core/cli/src/plugin/options.ts
ivomurrell: it's not that bad if you set GitHub to ignore the whitespace!

plugins/heroku/.toolkitrc.yml
ivomurrell: let's augment the doc generation script in a separate PR?

PR #640, opened 2024-04-30
fix: require package-json-hook plugin in plugins that use PackageJson hook

apaleslimghost:
ronseal

### DIFF
--- a/plugins/husky-npm/package.json
+++ b/plugins/husky-npm/package.json
@@ -14,8 +14,8 @@
     "tslib": "^2.3.1"
   },
   "peerDependencies": {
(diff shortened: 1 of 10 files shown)


plugins/husky-npm/package.json
```diff
@@ -14,8 +14,8 @@
     "tslib": "^2.3.1"
   },
   "peerDependencies": {
-    "husky": "4.x",
-    "dotcom-tool-kit": "3.x"
+    "dotcom-tool-kit": "3.x",
+    "husky": "4.x"
```
apaleslimghost, line 18: 🤷🏻‍♀️

PR #648, opened 2024-05-09
fix: pass through unparsed merged options if no schema

apaleslimghost:
to allow custom plugins to use task/command options


ivomurrell:
oops, accidentally re-implemented in #652 😄

PR #669, opened 2024-07-22
Don't run review jobs on tagged releases

ivomurrell:
# Description

We are already running staging jobs on tagged releases, so we don't need to run review jobs too. In addition, the `HerokuReview` task will fail as it can't find a branch to associate the review app with. For an example of why this is a problem, see next-router's CI [workflow](https://app.circleci.com/pipelines/github/Financial-Times/next-router/3073/workflows/8deeedd2-4e71-4916-86bd-99c50e2f64d8) with its unnecessary `tool-kit/deploy-review-node` branch.

# Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,13 +47,13 @@
     },
     "core/cli": {
       "name": "dotcom-tool-kit",
-      "version": "3.4.0",
+      "version": "3.5.1",
       "license": "MIT",
       "dependencies": {
         "@dotcom-tool-kit/error": "^3.2.0",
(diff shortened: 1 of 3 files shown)


ivomurrell:
> Do we still want to deploy to staging and production on tagged releases?

You're right, [we didn't use to](https://github.com/Financial-Times/next-router/blob/168cc8d9111a1ffcf0715fe655cc2845eeeac38d/.circleci/config.yml#L258-L278)! I'll update this PR.

PR #679, opened 2024-07-30
CPP-2161 proof of concept parallel tasks

apaleslimghost:
implements parallel task running in userland using a task with options, which looks like:

```yml
commands:
  run:local:
    - Parallel:
        tasks:
          - Node:
              entry: server/index.js
          - Webpack:
              watch: true
```

#### still to do

- [ ] tests
- [ ] update tasks that run in background to wait
- [ ] use parallel for preset plugins where they're expecting tasks to background


plugins/parallel/test/tasks/parallel.test.ts
```diff
@@ -0,0 +1,202 @@
...
+
+        await expect(task.run({ command: 'test:command', cwd: __dirname, config })).rejects
+          .toMatchInlineSnapshot(`
+          "ToolKitError" "error from task 1" {
+            "details": undefined,
+          }
+        `)
+
+        expect(stopMock).toBeCalledTimes(3)
+        expect(stopMock.mock.contexts).toEqual(task.taskInstances)
```
apaleslimghost, line 198: maybe the neatest test i've ever written ngl

plugins/parallel/readme.md
```diff
@@ -0,0 +1,69 @@
+# @dotcom-tool-kit/parallel
+
+This plugin allows you to run Tool Kit tasks in parallel.
```
olusoladavid, line 3:
```suggestion
This plugin allows you to run Tool Kit tasks in parallel. By default, Tool Kit tasks run sequentially in the order they are specified in a command.
```

plugins/parallel/readme.md
```diff
@@ -0,0 +1,69 @@
...
+	- '@dotcom-tool-kit/parallel'
+```
+
+## Usage
+
+To run tasks in parallel, pass them in as the `tasks` option to the [`Parallel` task](#parallel).
+
+`Parallel` is intended for running long-running tasks alongside each other, such as running your server with the `Node` task at the same time as bundling your code with the `Webpack` task in watch mode.
+
+If any of the tasks throw an error, `Parallel` will throw an error, and attempt to stop the other running tasks. This ensures that either all of the tasks you expect to be running are actually running, or that Tool Kit has exited and printed an error message.
```
olusoladavid, line 26: **question:** Is this still the default behavior or do we need to update this part?

plugins/parallel/src/tasks/parallel.ts
```diff
@@ -0,0 +1,127 @@
...
+      `  - ${styles.task(task.task)} ${styles.dim(
+        `(with options ${styles.code(JSON.stringify(task.options))})`
+      )}`
+  )
+  .join('\n')}
+`)
+
+    // HACK:KB:20250619 loadTasks expects config to be mutable, in TaskRunContext it's readonly, just cast it idc
+    this.taskInstances = (
+      await loadTasks(this.logger, tasks, context.config as WritableDeep<ValidConfig>)
```
olusoladavid, line 72: **question:** Is it possible to create a mutable copy of the config here instead (eg with `structuredClone`) as opposed to casting the readonly object to be mutable?

plugins/parallel/src/tasks/parallel.ts
```diff
@@ -0,0 +1,127 @@
...
+        )
+
+        if (errors.length > 0) {
+          handleTaskErrors(errors, context.command)
+        }
+
+        break
+      }
+      default: {
+        this.assertUnreachable('Parallel.options.onError', this.options.onError)
```
olusoladavid, line 106: **question:** would this be triggered if the user passed in an invalid value to the `onError` option?

plugins/parallel/src/tasks/parallel.ts
```diff
@@ -0,0 +1,127 @@
...
+      }
+    }
+  }
+
+  async stop() {
+    await Promise.all(
+      this.taskInstances.map((task) =>
+        task.stop().catch((error) => {
```
olusoladavid, line 114: **question:** If a non-persistent long-running task is in the `Parallel` task group being executed, will Parallel have to wait for the task in each `onError` case before fully exiting?

plugins/parallel/test/tasks/parallel.test.ts
olusoladavid, line 1: **question:** will the tests cover for persistent long-running processes too? or do we need a specific test case for that? It might also make sense to have a test case for a mix of process types and how logging is handled in that case where non-persistent processes may overlap?

plugins/webpack/src/tasks/webpack.ts
```diff
@@ -39,8 +41,15 @@ export default class Webpack extends Task<{ task: typeof WebpackSchema }> {
...
 
-    const child = fork(webpackCLIPath, args, { silent: true, execArgv, cwd })
-    hookFork(this.logger, 'webpack', child)
-    return waitOnExit('webpack', child)
+    this.child = fork(webpackCLIPath, args, { silent: true, execArgv, cwd })
+    hookFork(this.logger, 'webpack', this.child)
+    return waitOnExit('webpack', this.child)
+  }
+
+  async stop() {
```
olusoladavid, line 49: **question:** how will plugin authors know when to implement the `stop()` method? do we need to document this?

plugins/monorepo/src/tasks/workspace-command.ts
```diff
@@ -3,7 +3,7 @@ import { styles } from '@dotcom-tool-kit/logger'
 import { minimatch } from 'minimatch'
 import pluralize from 'pluralize'
 import { z } from 'zod'
-import { runTasksFromConfig } from 'dotcom-tool-kit/lib/tasks'
+import { runCommandsFromConfig } from 'dotcom-tool-kit/lib/tasks'
```
olusoladavid, line 6: **praise:** thank you for refactoring and adding clarity to this

olusoladavid (commented):
Great work! Haven't had the chance to fully test the new UX but the code approach is solid 🔥 . Thanks for putting together. I added some questions in the meantime

plugins/parallel/test/tasks/parallel.test.ts
apaleslimghost: good question! i'm going to add those tests in #998, will be easier to test once we've got a good example of how it should work in practice

plugins/parallel/src/tasks/parallel.ts
apaleslimghost: it turns out it's possible to make this config arg in `loadTasks` `ReadonlyDeep` as well, so i've done that

plugins/parallel/readme.md
apaleslimghost: good catch, it's not the default any more. the current behaviour is better documented in the options section so i've removed this part

plugins/webpack/src/tasks/webpack.ts
apaleslimghost: i've documented this in [`extending-tool-kit.md`](https://github.com/Financial-Times/dotcom-tool-kit/pull/679/files#diff-e2557f5acc2fc22ebb8900a7fc63005b9827d8eb9ce3623cecb096b33bcd2a65)

plugins/parallel/src/tasks/parallel.ts
olusoladavid: For the case of `onError: stop-all`, if there is a long running task (could be finite but still takes a long time to exit) that doesn't implement the `stop()` method, would `Parallel` have to wait for this task to complete before it then renders the original error?

PR #629, opened 2024-04-23
Partially autogenerate plugin readmes from schemas

apaleslimghost:
i'd mostly just review this by looking at the rich diff of the readmes. also maybe look at the script.

this PR also makes some drive-by changes to the `UploadAssetsToS3` task to remove the legacy environment variable handling.

PR #683, opened 2024-09-02
Remove deprecated plugins and code before major release

ivomurrell:
# Description

Now that we're gearing up to do a major release of the `next` branch, let's incorporate some cleanup of old deprecated code which we've kept around to avoid breaking changes. 

# Checklist:


plugins/circleci-deploy/readme.md
```diff
@@ -35,12 +35,12 @@ See [manually adding jobs to your CircleCI config](https://github.com/Financial-
 
 ## Hooks
```
apaleslimghost, line 36: **thought**: we should update these readmes! didn't get caught in #629

plugins/heroku/test/promoteStagingToProduction.test.ts
```diff
@@ -3,9 +3,8 @@ import { promoteStagingToProduction } from '../src/promoteStagingToProduction'
 import heroku from '../src/herokuClient'
 import { gtg } from '../src/gtg'
 import winston, { Logger } from 'winston'
-import { setAppConfigVars } from '../src/setConfigVars'
 
-const logger = (winston as unknown) as Logger
+const logger = winston as unknown as Logger
```
apaleslimghost, line 7: **question**: i'm sure this keeps flipflopping, what's causing that?

apaleslimghost (approved):
_mario voice_ buh-bye!

plugins/heroku/test/promoteStagingToProduction.test.ts
ivomurrell: looks like it's just the heroku tests that still have parentheses in this casting at this point, i expect we're just slowly applying prettier throughout the source as we make changes. (i still need to fix this test logging, winston doesn't like you just using the root logger without any transports configured and will print loads of complaints to stderr as soon as you switch off `--silent`)

PR #687, opened 2024-09-10
feat(commitlint): add a Commitlint plugin

rowanmanning:
# Description

This adds a Commitlint plugin, copied from a mix of Content Pipeline and various Reliability repos - we're now maintaining this same code in [eight different places](https://github.com/search?q=org%3AFinancial-Times+Commitlint+extends+Task+language%3AJavaScript&type=code&l=JavaScript).

I went for a broad range on the peer dependency (16–19) because this captures everything we use across the FT and the command line API we're using doesn't change between these versions.

I haven't added tests (yet?) because it seemed like quite a lot of setup to test a relatively small amount of code. I'm happy to give it a go but probs want direction on what to test. I was originally thinking "Do I need to spin up a GitHub repo in a temporary folder and try committing to verify that this fails?". Going by your other tests they all seem more end-to-end than unit and I couldn't find a lot of mocking.

# Checklist:


plugins/commitlint/readme.md
```diff
@@ -0,0 +1,20 @@
...
+```
+
+And add it to your repo's `.toolkitrc.yml`:
+
+```yml
+plugins:
+  - '@dotcom-tool-kit/commitlint'
+```
+
+By default the plugin will run on the `git:commitmsg` command. There are no further configurations in Tool Kit, [use one of the config file locations that Commitlint supports](https://commitlint.js.org/reference/configuration).
```
apaleslimghost, line 20: **suggestion**: although the task has no options, you should still add an [empty schema](https://github.com/Financial-Times/dotcom-tool-kit/blob/main/lib/schemas/src/tasks.ts#L23) in `schemas/tasks.ts` with a description for the task, add `<!-- begin autogenerated docs -->` to this readme, and run `scripts/generate-docs.js`

plugins/commitlint/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    },
+    {
+      "path": "../../lib/schemas"
+    }
+  ],
+  "include": ["src/**/*"],
+  "compilerOptions": {
+    "outDir": "lib",
+    "rootDir": "src",
+    "types": ["node", "mocha"]
```
apaleslimghost, line 18: **suggestion**: i don't think `mocha` is needed here?

```suggestion
    "types": ["node"]
```

plugins/commitlint/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+  "references": [
+    {
+      "path": "../../lib/logger"
+    },
+    {
+      "path": "../../lib/base"
+    },
+    {
+      "path": "../../lib/schemas"
+    }
```
apaleslimghost, line 12: **suggestion**: the reference to `schemas` isn't needed either

```suggestion
    }
```

plugins/commitlint/jest.config.js
apaleslimghost, line 1: **suggestion**: this can be deleted if we're not doing tests. i'm fine with no tests for this as doing an e2e test of the kind we're doing elsewhere would be a lot of overhead.

apaleslimghost (approved):
nice, thanks! a few small repo infrastructure type suggestions

plugins/commitlint/tsconfig.json
rowanmanning: Sorry copy pasta and this is the one file I forgot to check

plugins/commitlint/jest.config.js
rowanmanning: Ah I did wonder this but most of the other plugins that don't have tests still have this file and define a `test` script 😂 I left it in for consistency

apaleslimghost: oh huh fine

rowanmanning: If I remove this, what should `scripts.test` be set to?

apaleslimghost: the `test` script in the plugin's `package.json` is only used for local development for running the tests for that one package, so if it doesn't have tests you can remove it. the `cypress` plugin has it set to the [default `no test specified`](https://github.com/Financial-Times/dotcom-tool-kit/blob/main/plugins/cypress/package.json#L7) 😂

plugins/commitlint/readme.md
rowanmanning: Been faffing around with this and can't seem to get it to generate anything. I have the empty schema and nothing is changing when I generate docs. I tried logging stuff and it seems like `formatPluginSchemas` isn't returning anything

rowanmanning: Oh nevermind need to add it to the task

PR #695, opened 2024-10-03
fix(cli): add missing dependency on @dotcom-tool-kit/state

apaleslimghost:
noticed by @rowanmanning. in practice, most repos install a plugin that depends on `state`, which is why we haven't noticed this until now.

we should probably set up `eslint-plugin-import` to catch this!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,16 +50,17 @@
     },
     "core/cli": {
       "name": "dotcom-tool-kit",
-      "version": "4.0.1",
+      "version": "4.0.2",
       "license": "MIT",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -50,16 +50,17 @@
     },
     "core/cli": {
       "name": "dotcom-tool-kit",
-      "version": "4.0.0",
+      "version": "4.0.1",
```
apaleslimghost, line 53: release-please my beloathed

ivomurrell (approved):

PR #712, opened 2024-11-20
feat(schemas): add validation warnings for plugin options that moved to task options

apaleslimghost:
when migrating from v3, Tool Kit will happily accept these options, even though they're not present in the plugin schema (because it's not strict). this provides helpful error messages to instruct the user that these options need to be moved (which will make the migration guide easier to follow).

### DIFF
--- a/lib/schemas/src/moved-plugin-options.ts
+++ b/lib/schemas/src/moved-plugin-options.ts
@@ -1,0 +1,10 @@
+import { styles } from '@dotcom-tool-kit/logger'
+
+export const movedPluginOptions = <T extends Record<string, unknown>>(option: string, task: string, newName = option) => [
+	(options: T) => !(option in options),
+	{
+	  message: `the option ${styles.code(option)} has moved to ${styles.code(
+					  `options.tasks.${styles.task(task)}.${newName}`
+					)}`
+	}
+ ] as const
--- a/lib/schemas/src/prompts.ts
+++ b/lib/schemas/src/prompts.ts
@@ -19,14 +19,19 @@ export type SchemaPromptGenerator<T> = (
   // bizOpsSystem will be undefined if project is not a system
   bizOpsSystem?: BizOpsSystem
 ) => Promise<T | undefined>
+
+type InferZodShape<T> = T extends z.ZodObject<infer Shape>
+  ? Shape
+  : // we use .refine to warn about moved plugin options but each call to .refine
+  // wraps the Zod schema type in a ZodEffects which we need to unwrap first
+  T extends z.ZodEffects<infer Unwrapped>
+  ? InferZodShape<Unwrapped>
+  : never
+
 // This type defines an interface you can use to export prompt generators. The
 // `T` type parameter should be the type of your `Schema` object, and it will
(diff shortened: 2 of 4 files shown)


lib/schemas/src/prompts.ts
```diff
@@ -1,6 +1,6 @@
 import type prompts from 'prompts'
 import type { Logger } from 'winston'
-import type { z } from 'zod'
+import { object, type z } from 'zod'
```
ivomurrell, line 3: **nitpick**: `object` isn't used

```suggestion
import type { z } from 'zod'
```

lib/schemas/src/prompts.ts
```diff
@@ -19,11 +19,19 @@ export type SchemaPromptGenerator<T> = (
   // bizOpsSystem will be undefined if project is not a system
   bizOpsSystem?: BizOpsSystem
 ) => Promise<T | undefined>
+
+// HACK:KB:20241119 we use .refine to warn about moved plugin options but each call
+// to .refine wraps the Zod schema type in a ZodEffects which breaks the infer below
+type NestedZodEffects<T extends z.ZodTypeAny> =
```
ivomurrell, line 25: **suggestion**: we can use a recursive `infer` here to allow for an arbitrary number of `.refine` calls. we can also bundle the `infer` for the inner object shape here which allows us to simplify the `PromptGenerators` definition because TypeScript seems to infer the type constraints better!

```ts
type InferZodShape<T> = T extends z.ZodObject<infer Shape>
  ? Shape
  : // we use .refine to warn about moved plugin options but each call to .refine
  // wraps the Zod schema type in a ZodEffects which we need to unwrap first
  T extends z.ZodEffects<infer Unwrapped>
  ? InferZodShape<Unwrapped>
  : never

// This type defines an interface you can use to export prompt generators. The
// `T` type parameter should be the type of your `Schema` object, and it will
// be mapped into a partial object of `SchemaPromptGenerator` functions with
// all their return types set to the output type of each option schema.
export type PromptGenerators<T> = {
  [option in keyof InferZodShape<T>]?: SchemaPromptGenerator<z.output<InferZodShape<T>[option]>>
}
```

lib/schemas/src/moved-plugin-options.ts
```diff
@@ -0,0 +1,10 @@
+import { styles } from '@dotcom-tool-kit/logger'
+
+export const movedPluginOptions = <T extends Record<string, unknown>>(option: string, task: string, newName = option) => [
```
ivomurrell, line 3: **praise**: this is very nice

ivomurrell (approved):
**praise**: this is a great idea and a nice implementation!

I got nerd-sniped by trying to improve the type inference

lib/schemas/src/prompts.ts
apaleslimghost: dunno how that got there!

lib/schemas/src/prompts.ts
apaleslimghost: amended the original commit with this, thanks!

PR #713, opened 2024-11-21
V4 migration guide

apaleslimghost:
ronseal

### DIFF
--- a/docs/migration-guides/v4.md
+++ b/docs/migration-guides/v4.md
@@ -15,6 +15,8 @@ npm query ":root > [name*=dotcom-tool-kit]" \
 | xargs npm install
 ```
 
+If the `npm install` fails with an error suggesting the use of `--force`, edit the command to call `xargs npm install --force` last instead and run again. This should fix npm's version resolution and shouldn't result in any version errors.
+
 <details><summary>what</summary>
 
 1. `npm query ":root > [name*=dotcom-tool-kit]"`: get all packages installed as a direct dependency of your repo with `dotcom-tool-kit` in their name
--- a/docs/migration-guides/v4.md
+++ b/docs/migration-guides/v4.md
@@ -198,17 +198,16 @@ Add the `version` to your plugin's `.toolkitrc.yml` (most plugins' `.toolkitrc.y
 
 ### 2.2. Plugins with custom tasks
 
-> [!TIP]
-> Check to see if your custom task can now be provided by a built-in plugin. For example, if you're defining a custom task so you can run the same task multiple times with different options, this can now be done via command-specific task options:
->
-> ```yml
-> commands:
->   test:local:
->     - Jest
->   test:e2e-local:
->     - Jest:
->         configFile: jest.e2e.config.js
-> ```
+First, check to see if your custom task can now be provided by a built-in plugin. For example, if you're defining a custom task so you can run the same task multiple times with different options, this can now be done via command-specific task options:
+
+```yml
+commands:
+  test:local:
+    - Jest
+  test:e2e-local:
+    - Jest:
+        configFile: jest.e2e.config.js
+```
 
 #### 2.2.1. Import base classes from `@dotcom-tool-kit/base`
 
--- a/docs/migration-guides/v4.md
+++ b/docs/migration-guides/v4.md
@@ -38,7 +38,7 @@ npm install dotcom-tool-kit@latest @dotcom-tool-kit/jest@latest @dotcom-tool-kit
 ### 1.1. Rename `hooks` to `commands`
 
 > [!NOTE]
-> **Why?** In previous versions of Tool Kit, hooks both defined names of commands to run, and managed configuration files. In Tool Kit v4, hooks only manage configuration files, and commands are defined in the Tool Kit configuration files, not in plugin code.
+> **Why?** A Tool Kit "command" is a name for a group of tasks to run, like `test:local`. In previous versions of Tool Kit, hooks both defined names of commands to run, and managed configuration files. In Tool Kit v4, hooks only manage configuration files, and commands are defined in the Tool Kit configuration files, not in plugin code.
 
 In your `.toolkitrc.yml`, you may have a `hooks` object, which specifies which Tool Kit tasks to run for commands. **Rename this object to `commands`**:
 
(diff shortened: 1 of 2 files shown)


docs/migration-guides/v4.md
```diff
@@ -0,0 +1,327 @@
...
+> [!NOTE]
+> This migration will take 30m-1hr. If at any point you get stuck or something unexpected happens, [reach out to the Platforms team](https://financialtimes.enterprise.slack.com/archives/C02TRE2V2Q1) and we'll be happy to help.
+
+## 0. Update to latest versions of Tool Kit packages
+
+Update `dotcom-tool-kit` and all the `@dotcom-tool-kit/*` packages you have installed to their latest versions. This can be done with the following command (requires [`jq`](https://jqlang.github.io/jq/)):
+
+```sh
+npm query ":root > [name*=dotcom-tool-kit]" \
+| jq '(.[] | .name + "@latest")' \
```
ivomurrell, line 14: **nitpick**: i think the parentheses are redundant

```suggestion
| jq '.[] | .name + "@latest"' \
```

docs/migration-guides/v4.md
```diff
@@ -0,0 +1,327 @@
...
+| xargs npm install
```
ivomurrell, line 15: **suggestion**: i've sometimes found that you need to call `npm install --force` instead (boo) might be worth calling that out

> If the `npm install` fails with an error suggesting the use of `--force`, edit the command to call `xargs npm install --force` last instead and run again. This should fix npm's version resolution and shouldn't result in any version errors.

idk something like that

docs/migration-guides/v4.md
```diff
@@ -0,0 +1,327 @@
...
+```
+
+<details><summary>what</summary>
```
ivomurrell, line 18: **praise**: great explainer

docs/migration-guides/v4.md
```diff
@@ -0,0 +1,396 @@
...
+		};
+	}
+}
+
+exports.hooks = {
+	'deploy:production': DeployProduction
+}
+```
+
+This adds a `tool-kit/deploy-production` job that depends on the `tool-kit/test-ci` job. To replicate this, add this configuration in your repo's `.toolkitrc.yml`:
```
apaleslimghost, line 365: i haven't tested this, but also there's literally [one use case and it's ours](https://github.com/Financial-Times/next-service-registry/blob/d31a6f4beb12d3b59dd7eba55bc1019cd3454925/toolkit/deploy-production/index.js#L5), so

docs/migration-guides/v4.md
```diff
@@ -0,0 +1,396 @@
...
+                  - 'test'
+                custom:
+                  filters:
+                    branches:
+                      only: main
+```
+
+### 2.5. Plugins with hooks doing other things
+
+You'll need to [talk to the Platforms team](https://financialtimes.enterprise.slack.com/archives/C02TRE2V2Q1) to get support with your migration.
```
apaleslimghost, line 388: again: nobody actually has any hooks doing other things. mostly i'm including this because looking like we're covering all the bases will give people more confidence

docs/migration-guides/v4.md
```diff
@@ -0,0 +1,327 @@
...
+(but without you having to manually type that out).
+
+</details>
+
+## 1. Make required changes to configuration
+
+### 1.1. Rename `hooks` to `commands`
+
+> [!NOTE]
+> **Why?** In previous versions of Tool Kit, hooks both defined names of commands to run, and managed configuration files. In Tool Kit v4, hooks only manage configuration files, and commands are defined in the Tool Kit configuration files, not in plugin code.
```
ivomurrell, line 39: **suggestion**: think here would be a good place to expand on what we mean by a command

```suggestion
> **Why?** In previous versions of Tool Kit, hooks both defined names of commands to run, and managed configuration files. In Tool Kit v4, hooks only manage configuration files, and commands are defined in the Tool Kit configuration files, not in plugin code. Commands comprise of an explicitly ordered set of tasks to run.
```

docs/migration-guides/v4.md
```diff
@@ -0,0 +1,398 @@
...
+
+Add the `version` to your plugin's `.toolkitrc.yml` (most plugins' `.toolkitrc.yml` files will be empty before this):
+
+```diff
++version: 2
+```
+
+### 2.2. Plugins with custom tasks
+
+> [!TIP]
```
ivomurrell, line 199: **todo**: this tip isn't rendering properly because it's quoted

ivomurrell (approved):
this is so so good, thank you kara!!!

docs/migration-guides/v4.md
apaleslimghost: so the previous version of the command that's been passed around (the one with `xargs -I{} npm install {}@latest` ran the `npm install`s one by one which is why `--force` was needed. this one should(!) work fine as it does the install in one step. do you think we still need the note?

docs/migration-guides/v4.md
apaleslimghost: wdym "because it's quoted"? the blockquote syntax is how you do the tip blocks (i mean, yeah, this one isn't working, but idk why, maybe the code block? it looked fine in vs code preview 😕)

docs/migration-guides/v4.md
ivomurrell: hm i've been using the correct `xargs npm install` version for months now and still need to use `xargs npm install --force` sometimes yeah (verified by checking my terminal history, can see uses from 6 months ago)

apaleslimghost: okay! i'll add the warning

docs/migration-guides/v4.md
ivomurrell: ah sorry i just assumed that was why but you know more than me!

apaleslimghost: lmao it's because it's nested in a `<details>` every other bit of markdown syntax is working fine 🥴

alexmuller:
👍 This is great. It's a lot to absorb but I think the callouts, details/summary and syntax highlighted diffs make it easy to follow.

docs/migration-guides/v4.md
```diff
@@ -0,0 +1,399 @@
+# Migrating to Tool Kit v4
```
CyntiBinti, line 1: suggestion: helpful to have a contents list with hyperlinks to the subheadings for easier navigation through the doc

CyntiBinti (approved):
Nice one! 🙌🏽

GlynnPhillips:
@apaleslimghost this is great. I especially echo what Alex said about the callouts/notes/summaries making this much easier to consume given how much there there is to take in

PR #714, opened 2024-11-21
CPP-2275: resolve a plugin's .toolkitrc.yml

apaleslimghost:
resolving package.json works fine for built-in plugins, but for custom plugins this effectively means they need to have an index.js, which most won't any more. resolve the .toolkitrc.yml, which every built-in and custom plugin must have

### DIFF
--- a/core/cli/src/plugin/resolve-root.ts
+++ b/core/cli/src/plugin/resolve-root.ts
@@ -1,11 +1,10 @@
 import path from "path"
 
 export function resolveRoot(id: string, root: string): string {
-  const isPath = id.startsWith('.') || id.startsWith('/')
-  // resolve the package.json of a plugin as many plugins don't have valid
+  // resolve the .toolkitrc.yml of a plugin as many plugins don't have valid
   // entrypoints now that we're intending their tasks/hooks to be loaded via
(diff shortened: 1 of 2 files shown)


ivomurrell (changes requested):
**issue**: I don't _think_ this will work as the `resolve` function will still be using the same `require` resolution logic which will be looking for modules so if a plugin doesn't have a `package.json` it will fail to resolve the path.

is this change actually required? this function is for resolving the path of the root plugin – custom plugins will likely never be the root and will always be imported by a `.toolkitrc.yml` at the project's root which will have a `package.json`

apaleslimghost:
> this function is for resolving the path of the root plugin

not true! it's for resolving any plugin's root, it's called by `loadPlugin`

> will still be using the same require resolution logic which will be looking for modules so if a plugin doesn't have a `package.json` it will fail to resolve the path

nuh uh

Screenshot 2024-11-21 at 1 59 46 PM

> is this change actually required?

without this we'll try to resolve a custom plugin to its root folder, which will then try to resolve `index.js` or `package.json`, which will usually fail

i do need to test this properly! and this area isn't very well tested tbh

PR #717, opened 2024-11-26
preserve relative paths in resolveRoot to fix custom plugin loading

apaleslimghost:
and do some cursed things to work around Jest's different behaviour for `require.resolve` in the tests

### DIFF
--- a/core/cli/src/plugin/resolve-root.ts
+++ b/core/cli/src/plugin/resolve-root.ts
@@ -1,5 +1,7 @@
 import path from "path"
 import { resolve } from "./require-resolve"
+import { ToolKitError } from "@dotcom-tool-kit/error"
+import { styles } from "@dotcom-tool-kit/logger"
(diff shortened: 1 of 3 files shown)


core/cli/src/plugin/resolve-root.ts
```diff
@@ -1,10 +1,26 @@
 import path from "path"
+import { resolve } from "./require-resolve"
+import { ToolKitError } from "@dotcom-tool-kit/error"
+import { styles } from "@dotcom-tool-kit/logger"
 
 export function resolveRoot(id: string, root: string): string {
+  const isRelative = id.startsWith('./')
```
apaleslimghost, line 7: previously we were checking for absolute paths here as well, but they're preserved by `path.join`, it's only `./` relative paths that aren't

ivomurrell (approved):

PR #719, opened 2024-11-27
Allow CircleCI workflow job options to be overridden

ivomurrell:
# Description

Previously [jobs](https://circleci.com/docs/configuration-reference/#jobs-in-workflow) in CircleCI workflows (i.e., where you define which jobs a workflow runs and which custom parameters to use) could not be overridden by other plugins as they were a deeper part of the hook options that the merging logic didn't reason about.

This allows us to move the Cloudsmith changes in #710 into a separate plugin. 

# Checklist:


lib/schemas/src/hooks/circleci.ts
```diff
@@ -3,65 +3,79 @@ import { z } from 'zod'
...
-  name: z.string(),
-  image: z.string()
-})
+export const CircleCiExecutor = z
+  .object({
+    name: z.string(),
+    image: z.string()
+  })
+  .partial()
+  .required({ name: true })
```
apaleslimghost, line 12: **question**: what's the difference between this and doing `image: z.string().optional()`?

plugins/circleci/src/circleci-config.ts
```diff
@@ -226,6 +227,28 @@ const customOptionsOverlap = (
...
+    const otherWorkflowJob = other.find(({ name }) => installationWorkflowJob.name === name)
+    return (
+      otherWorkflowJob &&
+      ((installationWorkflowJob.runOnRelease !== undefined &&
+        otherWorkflowJob.runOnRelease !== undefined &&
+        installationWorkflowJob.runOnRelease !== otherWorkflowJob.runOnRelease) ||
+        (installationWorkflowJob.splitIntoMatrix !== undefined &&
+          otherWorkflowJob.splitIntoMatrix !== undefined &&
+          installationWorkflowJob.splitIntoMatrix !== otherWorkflowJob.splitIntoMatrix) ||
+        customOptionsOverlap(installationWorkflowJob.custom, otherWorkflowJob.custom))
```
apaleslimghost, line 247: **nitpick**: don't really know how you could simplify this but it is _very_ dense right now! logic looks fine though?

apaleslimghost (approved):

plugins/circleci/src/circleci-config.ts
ivomurrell: yeaaah i should probably refactor this logic into a function like `areDefinedAndEqual` shouldn't i

lib/schemas/src/hooks/circleci.ts
ivomurrell: I want to preserve the semantics of the original schema – `image` is still required for the final merged schema, just not required in a single plugin's hook options. We don't do any validation on the final schema at the moment so I think it makes sense to call `.partial` directly on the schema but in the future they could be split out.

apaleslimghost: oh gotcha, i think i understand

PR #720, opened 2024-11-27
delete @dotcom-tool-kit/options

apaleslimghost:
having the options stored in a singleton isn't very maintainable, and isn't compatible with the monorepo plugin's isolated config objects for workspace packages.

this gets rid of the options singleton and `options` package, replacing it with explicitly passing the options down from the `config` where they're needed by things that aren't tasks or hooks.


core/cli/src/tasks.ts
```diff
@@ -97,7 +90,7 @@ export async function runTasksFromConfig(
     for (const task of tasks) {
       try {
         logger.info(styles.taskHeader(`running ${styles.task(task.id)} task`))
-        await task.run({ files, command, cwd: config.root })
+        await task.run({ files, command, cwd: config.root, config })
```
ivomurrell, line 93: **thought**: this is going to lead to future crimes i bet

ivomurrell (approved):

core/cli/src/tasks.ts
apaleslimghost: yeah that was my worry too. if you can think of another way of achieving this i'm all ears!

lib/base/src/hook.ts
```diff
@@ -5,14 +5,19 @@ import type { z } from 'zod'
 import type { Plugin } from '@dotcom-tool-kit/plugin'
 import { Conflict, isConflict } from '@dotcom-tool-kit/conflict'
 
+type Default<T, D> = T extends undefined ? D : T
```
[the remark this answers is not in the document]
apaleslimghost: can do. i'm surprised `type-fest` doesn't have this type!

core/cli/src/tasks.ts
ivomurrell: do you think we should `Object.freeze` the object before we pass it here to minimise the crime potential?

apaleslimghost: and make it `readonly` in the types

PR #721, opened 2024-12-02
Node watch

apaleslimghost:
Nodemon doesn't work nicely with the monorepo plugin, and is generally not well maintained. Node supports this option since v16.19 and v18.11 (albeit with an experimental warning) and imo we should be using built in platform features like this where possible instead of third party tools.


ivomurrell (approved):
nice simple addition, thanks!

PR #722, opened 2024-12-06
fix: add cypressImage to circleci schema

apaleslimghost:
if this isn't present in the schema, it isn't available in the parsed options, so `@toolkit/if-defined` doesn't see it.


ivomurrell:
I don't know why/if this was working before but this makes sense! Thanks for the comprehensive description!

ivomurrell (approved):

PR #726, opened 2024-12-10
fix: allow non-tool-kit-orb jobs to be specified in workflows

apaleslimghost:
currently we unconditionally prefix all jobs in workflows with `tool-kit/`, preventing people using jobs from other orbs such as Slack. with this PR, jobs that already include an orb prefix are output verbatim, as well as jobs that were added by the same hook installation.


ivomurrell (approved):
thanks!

PR #727, opened 2024-12-10
Actually output jobs

apaleslimghost:
the `jobs` option for the `CircleCI` hook doesn't actually do anything. make it do something.


plugins/circleci/src/circleci-config.ts
```diff
@@ -382,6 +396,33 @@ const generateJobs = (
...
+    ...(job.workspace?.attach ? [attachWorkspaceStep] : []),
+    {
+      run: {
+        name: job.name,
+        command: `npx dotcom-tool-kit ${job.command}`
+      }
+    },
+    ...(job.workspace?.persist ? [persistWorkspaceStep] : [])
+  ],
+  ...job.custom
```
ivomurrell, line 423: **thought**: this doesn't allow us to set custom steps in a job which might be useful; maybe it would be worth adding additional options to insert steps before and after the primary `npx dotcom-tool-kit` step?

ivomurrell (approved):
thanks for this!

plugins/circleci/src/circleci-config.ts
apaleslimghost: i thought about that but i couldn't think of a nice way of specifying that in the config. you can override all of the `steps` in `custom` and i figure the extra boilerplate you need if doing that is fine for now

ivomurrell: i was thinking of having `pre` and `post` fields which take an array of steps. think that's better than overriding the whole `steps` field in `custom` so that we're still generating _something_. doesn't need to be in this PR unless content pipeline might use it though!

apaleslimghost: maybe useful later but for now let's just do the basics

PR #732, opened 2024-12-30
Add support for Node 22 and remove npm engine field

joelcarr:
Adding support for Node 22 to all `dotcom-tool-kit` packages. Node 18 reaching is EOL in April. Node 22 is now set as a new LTS version. We will be encouraging all FT systems + packages to use Node 22.

This PR also drops the npm engine field ([CPP-2246](https://financialtimes.atlassian.net/browse/CPP-2246)). Customer Products are encouraged to use the version of npm that comes packaged with Node which is managed with volta.sh. The `dotcom-tool-kit` packages will support for the npm versions that is bundled with the supported node versions. This removes the debt of explicitly defining npm support.

[CPP-2246]: https://financialtimes.atlassian.net/browse/CPP-2246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/create-plugin.js
```diff
@@ -35,11 +35,10 @@ pkg.homepage = `https://github.com/financial-times/dotcom-tool-kit/tree/main/${d
 pkg.author = 'FT.com Platforms Team <platforms-team.customer-products@ft.com>'
 pkg.files = ['/lib', '.toolkitrc.yml']
 pkg.engines = {
-  node: '18.x || 20.x',
-  npm: '7.x || 8.x || 9.x'
+  node: '18.x || 20.x || 22.x'
```
ivomurrell, line 38: **praise**: it's always easy to forget to keep this up to date!

ivomurrell (approved):

PR #741, opened 2025-01-07
Add a Docker plugin

rowanmanning:
# Description

This adds a new low-level Docker plugin which is used to build and push images. This allows you to use the following commands in your `.toolkitrc.yml`:

```yml
plugins:
  - '@dotcom-tool-kit/cloudsmith'
  - '@dotcom-tool-kit/docker'
commands:
  deploy:staging:
    - DockerBuild
    - DockerPush
options:
  plugins:
    '@dotcom-tool-kit/cloudsmith':
      serviceAccount: cp-reliability-read-write # example - we have an account for each team
    '@dotcom-tool-kit/docker':
      images:
        web:
          definition: Dockerfile
          name: my-app
```

The plugin config (over task configs) is required because each app needs to be able to specify any number of different images, e.g.

```yml
options:
  plugins:
    '@dotcom-tool-kit/docker':
      images:
        web:
          definition: Dockerfile
          name: my-app
        worker:
          definition: worker/Dockerfile
          name: my-app-worker
```

We discussed this approach ahead of time but it's now simpler than our original proposal.

We've tried to stick with some sensible defaults, e.g. the `registry` and definition are defaulted to `docker.packages.ft.com/cp-container-registry` and `Dockerfile` respectively. This allows us to use a super minimal config for many apps:

```yml
options:
  plugins:
    '@dotcom-tool-kit/docker':
      images:
        web:
          name: my-app
```

# Checklist:


plugins/docker/src/tasks/push.ts
```diff
@@ -0,0 +1,64 @@
...
+            '--username',
+            auth.username,
+            '--password-stdin',
+            auth.registry
+          ])
+          childLogin.stdin.setDefaultEncoding('utf-8')
+          childLogin.stdin.write(auth.password)
+          childLogin.stdin.end()
+          hookFork(this.logger, 'dockerLogin', childLogin)
+          await waitOnExit('dockerLogin', childLogin)
```
ivomurrell, line 47: **nitpick**: we haven't had any subcommand process names before but i'd like to standardise on kebab casing for them if that's okay (à la `man git-commit`)

```suggestion
          hookFork(this.logger, 'docker-login', childLogin)
          await waitOnExit('docker-login', childLogin)
```

plugins/docker/src/build-image-name.ts
```diff
@@ -0,0 +1,6 @@
+import { join as joinPath } from 'node:path'
+
+export function buildImageName({ registry, name }: { registry: string; name: string }) {
+  const tag = process.env.CIRCLE_BUILD_NUM || `local-${process.env.USER || 'unknown'}`
```
ivomurrell, line 4: **question**: will these tags be visible in Cloudsmith? i think we could try to add a little more context to them (e.g., appending `-ci` for images build in CI, using semver for published versions, commit hashes for PR images etc.) we can make use of Tool Kit state to keep track of the current image tag if we're worried about deterministically using the same tag in both build and push tasks.

ivomurrell (approved):
no major complaints, this is a great plugin, thank you!

plugins/docker/src/tasks/push.ts
rowanmanning: Sounds good 👍 have updated

plugins/docker/src/build-image-name.ts
rowanmanning: They will be, yep. I feel like we should talk about this more broadly? We whacked the build number in there because it was something easily accessible but open to other alternatives. @alexmuller: thoughts?

plugins/docker/src/tasks/push.ts
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        const child = spawn('docker', ['push', tag])
```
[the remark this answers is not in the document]
rowanmanning: It's not a requirement but doing this could be a way around the next step of _deploying_ the built image. If we can rely on a `latest` tag being available then we don't need any context about the build to deploy. One potential issue with this, though, is if multiple builds kick off at similar times and the `latest` version we deploy wasn't the same one we built/tested in the current workflow.

Basically if we can safely use it for deployment then let's. Otherwise I think there's no need because Cloudsmith shows you the most recent published version in the interface etc.

plugins/docker/src/build-image-name.ts
alexmuller: We should have a chat about this! I don't love `SHA1` because when you're looking at 2 versions during an incident it's impossible to tell which one is newer. `TAG` would work but then we're requiring everyone to implement semver which doesn't make a lot of sense for a frontend app.

alexmuller: Or ISO8601 datetime?

ivomurrell: i assume cloudsmith will still keep the older image around when you push a new image with the same tag and all the images will be timestamped so that _shouldn't_ be a problem?

rowanmanning: ISO-time + (tag OR build-number)?

`2025-01-26T10:00:00Z-v1`

ivomurrell: ^ I assumed correctly. It will also remove the tag from the older image which is causing us some concern 😅 We'll investigate further tomorrow.

rowanmanning: OK for now I'm doing the following to make sure we've got a few pieces of info. We can talk more about naming of these when we bring it to a broader audience.

I'm adding _all_ of these tags if the data is present. On CI:

  * `ci-<buildnum>`
  * `git-<short-sha>`
  * `git-<git-tag-name>`

In local dev:

  * `local-<username>`

Sound OK @ivomurrell / @alexmuller ? It gets us to the point where we _have_ all the info.

plugins/docker/src/image-info.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export function buildImageName({ registry, name }: { registry: string; name: string }) {
+  return joinPath(registry, name)
+}
+
+// See https://docs.docker.com/reference/cli/docker/image/tag/#description
+// We're just doing some basic sanitization now to avoid git tags causing issues.
+// If the tag isn't valid then Docker will error so it's not high risk
+function sanitizeDockerTag(tag: string): string {
+  return tag.replace(/[^a-z0-9\._-]+/i, '')
```
ivomurrell, line 11: **todo**: this currently only replaces the first instance of a consecutive series unsanitary characters, the `g` flag is needed to replace all instances

```suggestion
  return tag.replace(/[^a-z0-9\._-]+/gi, '')
```

plugins/docker/src/tasks/build.ts
```diff
@@ -0,0 +1,40 @@
...
+import { ToolKitError } from '@dotcom-tool-kit/error'
+
+export default class DockerBuild extends Task<{
+  plugin: typeof DockerSchema
+}> {
+  async run() {
+    // Iterate over different image types like web, worker, etc
+    for (const [imageIdentifier, imageOptions] of Object.entries(this.pluginOptions.images)) {
+      const imageName = buildImageName(imageOptions)
+      const imageTags = [imageName, ...getImageTagsFromEnvironment(imageOptions)]
```
ivomurrell, line 15: **question**: i think including the bare `imageName` in the list of tags is equivalent to tagging the image with `` `${imageName}:latest` ``. was that intentional?

plugins/docker/src/image-info.ts
rowanmanning: Oops! Good catch, thanks

plugins/docker/src/tasks/build.ts
rowanmanning: Yep I noticed that. Can't hurt right? Unless you think `latest` is gonna be confusing/misleading

ivomurrell: Nope I think it's the right thing to do, just wanted to check since you weren't sure about using `latest` in the [other thread](https://github.com/Financial-Times/dotcom-tool-kit/pull/741#discussion_r1905782292).

PR #744, opened 2025-01-09
CPP-2322 fix circleci option defaults

apaleslimghost:
the `.partial()` on the hook option objects is overriding the `.default()` on their properties.
short of splitting the schema into an incoming options schema that's meant to be merged from
multiple sources, and an outgoing schema for the final hook options, this is the most
straightforward way of doing this.

### DIFF
--- a/plugins/circleci/test/circleci-config.test.ts
+++ b/plugins/circleci/test/circleci-config.test.ts
@@ -1,15 +1,17 @@
 import { type HookInstallation } from '@dotcom-tool-kit/base'
-import type {
-  CircleCiWorkflowJob,
-  CircleCiJob,
-  CircleCiOptions
+import {
+  type CircleCiWorkflowJob,
+  type CircleCiJob,
+  type CircleCiOptions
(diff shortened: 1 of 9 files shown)


plugins/circleci/test/circleci-config.test.ts
```diff
@@ -574,4 +602,22 @@ describe('CircleCI config hook', () => {
...
+
+      const hookInstallationsPromise = loadHookInstallations(logger, config).then((validated) =>
+        validated.unwrap('hooks were invalid')
+      )
+
+      await expect(hookInstallationsPromise).resolves.toEqual([expect.any(CircleCi)])
+
+      const installation = (await hookInstallationsPromise)[0]
+
+      expect(YAML.stringify(await installation.install())).toMatchSnapshot()
```
ivomurrell, line 620: **thought**: it might be useful to somehow validate the config in the future too (probably by shelling out to `circleci config validate`)

ivomurrell (approved):

PR #761, opened 2025-01-20
CPP-2063 Co-locate schema definitions with plugins

ivomurrell:
# Description

This will allow custom plugins to define their own option schemas, as well as decoupling plugins from the schemas package, which had the annoying property of causing a cascade of plugins to bump whenever a schema was updated for a single plugin.

Now we load task and hook options with the same entry point as the base classes are loaded from, looking for an exported `schema` object which is expected to be a Zod schema. Plugin options are defined in a separate entry point which is specified by the new `optionsSchema` setting.

In order to avoid any breaking changes we'll still fall back to schemas exported by the `schema` package if a schema isn't found in the plugin so that bumping the `cli` won't require bumping all plugins too. Note that the `schemas` package will no longer be maintained and any future schema updates will be done in the plugins themselves.

# Checklist:


docs/extending-tool-kit.md
```diff
@@ -134,3 +134,32 @@ commands:
...
+```
+
+Additionally, we can define an options schema to ensure that the options passed to task are as we expect them to be, with the correct names and correct types. We use [`zod`](https://zod.dev/) to define our schemas. For plugin options, you should export your schema as the default object in a separate file and specify its path as a top-level `optionsSchema` option in your `.toolkitrc.yml`. For tasks and hooks, you should export your schema as a `schema` object in the same file as you export your task/hook. We could define a `zod` schema for the Rollup task like so:
+
+```js
+const z = require('zod')
+
+const RollupSchema = z.object({ configPath: z.string() })
+module.exports.schema = RollupSchema
+```
```
apaleslimghost, line 165: **praise**: noice

core/cli/src/plugin/entry-point.ts
```diff
@@ -37,21 +39,62 @@ export async function importEntryPoint<T extends { name: string } & Omit<typeof
...
-    `entrypoint ${s.filepath(entryPoint.modulePath)} in plugin ${s.plugin(
-      entryPoint.plugin.id
-    )} does not export a class as its ${s.code('default')} export or ${s.code('module.exports')}`
-  ])
+// the subclasses of Base have different constructor signatures so we need to omit
+// the constructor from the type bound here so you can actually pass in a subclass
+export async function importEntryPoint<T extends { name: string } & Omit<typeof Base, 'new'>>(
+  type: T,
+  entryPoint: EntryPoint
+): Promise<Validated<{ entryPoint: T; schema?: z.ZodSchema }>> {
```
apaleslimghost, line 67: **suggestion**: i think this return type is a little confusing. "entry point" is used to refer to the type that contains a reference to a plugin and a path to import, but here it's also the thing that's actually being imported. not sure what i'd suggest as a better property name though (`defaultExport` maybe? not really keen on that though)

apaleslimghost (approved):
fantastic! a few nonblocking suggestions

plugins/babel/package.json
```diff
@@ -20,7 +20,8 @@
     "@dotcom-tool-kit/error": "^4.1.0",
     "@dotcom-tool-kit/logger": "^4.1.1",
     "fast-glob": "^3.2.11",
-    "tslib": "^2.3.1"
+    "tslib": "^2.3.1",
+    "zod": "^3.24.1"
```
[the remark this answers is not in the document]
ivomurrell: I'm not going to do this yet – the only issue i've found is with the type checking (TypeScript complains that a plugin's options don't properly extend `z.ZodTypeAny` if the versions don't match) and i can't think of any runtime issues that we may run into as long as the major version of `zod` is the same. it seems like overkill to float a developer problem up to require our version of `zod` as a peer dependency in the consumer's project.

using peer dependencies _would_ make sure that you couldn't install a plugin which uses a future `v4` of `zod` along with an older core using `v3`, but bumping the major zod version would be a breaking change either way so I don't think that makes a difference as we already have a peer dependency on `dotcom-tool-kit`.

PR #762, opened 2025-01-21
Store pushed Docker images in state

rowanmanning:
We will need the built image names and tags in order to deploy via Hako. This involved a bit of refactoring of the tag generation so that we weren't duplicating stuff. We're using the CI build number for deploys for now but this can change with no issue later if we change our minds.

### DIFF
--- a/lib/state/src/index.ts
+++ b/lib/state/src/index.ts
@@ -6,12 +6,19 @@ const stateDir = target ? path.join(target, '.toolkitstate') : '.toolkitstate'
 
 type InstallState = Record<string, string>
 
-interface CIState {
+interface DockerImage {
+  fullyQualifiedName: string
+  name: string
+  tag: string
+}
+
+export interface CIState {
   repo: string
   branch: string
   version: string
   tag: string
   buildNumber: string
+  pushedImages: DockerImage[]
 }
 
 export interface ReviewState {
--- a/plugins/docker/src/image-info.ts
+++ b/plugins/docker/src/image-info.ts
@@ -1,5 +1,6 @@
-import { readState } from '@dotcom-tool-kit/state'
+import { CIState } from '@dotcom-tool-kit/state'
 import { join as joinPath } from 'node:path'
+import { randomInt } from 'node:crypto'
 
 export function buildImageName({ registry, name }: { registry: string; name: string }) {
   return joinPath(registry, name)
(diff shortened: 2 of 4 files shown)


plugins/docker/src/tasks/push.ts
```diff
@@ -60,5 +70,10 @@ export default class DockerPush extends Task<{
         }
       }
     }
+
+    // TODO either update "ci" types or add a new "docker" state??
+    writeState('ci', {
+      pushedImages
+    });
```
rowanmanning, line 77: We're not sure where we should be putting this info. We were originally thinking `ci` but maybe it makes more sense to add a new `docker` state?

Either way we'll need to update the types in `@dotcom-tool-kit/state`. Thoughts @apaleslimghost / @ivomurrell ?

apaleslimghost: state is kind of a mess tbh, so i wouldn't worry too much about whether this is the "right" bit of state to put it in. `ci` is fine.

plugins/docker/src/image-info.ts
```diff
@@ -12,38 +13,51 @@ function sanitizeDockerTag(tag: string): string {
...
-    tags.push(`branch-${branchName}`)
+  if (ciState?.branch) {
+    tags.gitBranch = `branch-${ciState.branch}`
   }
 
+  return tags
+}
+
+export function getDeployTag(): string {
+  return generateImageTags().buildNumber
```
ivomurrell, line 50: **suggestion**: maybe the `buildNumber` generation logic could go into this function instead so that we don't unnecessarily compute the other fields and also can skip the crypto call if we're in CI? i also like how this would make the call to `readState` the caller's concern which could help avoiding repeated implicit calls.

```suggestion
export function getDeployTag(ciState: CIState | null): string {
  return  ciState?.buildNumber ? `ci-${ciState.buildNumber}` : `local-${process.env.USER || 'unknown'}-${randomInt(1000, 9999)}`
```

lib/state/src/index.ts
```diff
@@ -6,12 +6,19 @@ const stateDir = target ? path.join(target, '.toolkitstate') : '.toolkitstate'
...
+  tag: string
+}
+
 interface CIState {
   repo: string
   branch: string
   version: string
   tag: string
   buildNumber: string
+  pushedImages: DockerImage[]
```
ivomurrell, line 21: **thought**: why are none of the fields in these state objects partial??

ivomurrell (approved):

plugins/docker/src/image-info.ts
rowanmanning: Happy to do this but wanted to not duplicate the logic for generating the deploy tag, maybe moving that into a separate function as well? or calling `getDeployTag` from inside `generateImageTags` as well?

ivomurrell: Yeah I was imagining the latter!

rowanmanning: Done 👍 does this look OK? Two things:

* Tweaked the naming a bit so that it's always called the deploy tag
* Right now `CIState` isn't exported from `@dotcom-tool-kit/state` so I used `State['ci']`. I avoiding exporting the interface this because I didn't want this to trigger a release of _every_ plugin but can do if you're OK with that, just avoiding the noise

rowanmanning: Oh lol wait, we're editing `state` anyway! Will update if you'd prefer us to just export `CIState` now

PR #763, opened 2025-01-23
fix: allow building ARM64 images in AMD CI jobs

rowanmanning:
# Description

You can't use the default `docker build` to build multi-arch images, this is something we need if we want to build ARM-based images in CircleCI. The alternative would be to move _all_ our CircleCI jobs to ARM which is more expensive and a bunch of work.

I tested this on our migration canary:
* [Here's the relevant code that ran successfully](https://github.com/Financial-Times/cp-docker-migration-canary/commit/7108fc6003279a3eb5b5a0a4c5c14f805fc0bfaf)
* [Here's the built image, note that it's ARM](https://cloudsmith.io/~financial-times/repos/cp-container-registry/packages/detail/docker/cp-docker-migration-canary/1ac7d058f41dd8d820f00fb50bd9f14611c981c6f465d5cde772edc04f2afd57/a=arm64;xpo=linux/)

# Checklist:

### DIFF
--- a/plugins/docker/src/tasks/build.ts
+++ b/plugins/docker/src/tasks/build.ts
@@ -15,8 +15,23 @@ export default class DockerBuild extends Task<{
       const imageTags = [imageName, ...getImageTagsFromEnvironment(imageOptions)]
       this.logger.info(`Building image "${imageIdentifier}": with tags ${imageTags.join(', ')}`)
       try {
-        const child = spawn('docker', [
+        // We need to create a builder so that we're not limited to
+        // building AMD64 images due to the platform our CircleCI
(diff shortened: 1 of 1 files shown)


CyntiBinti (approved):

ivomurrell (approved):
**praise**: looks good to me and great that this is possible!

**question(non-blocking)**: from what i can gather, this will mean that we'll be running [QEMU](https://docs.docker.com/build/building/multi-platform/#qemu) when building our docker images (i.e., doing 1 million syscalls when doing an `npm install`) – have you seen any performance impact/effect on CI times when doing this cross-compilation?

PR #764, opened 2025-01-24
feat: widen semver range for mocha

rowanmanning:
# Description

mocha v11 has only one breaking change - dropping support for Node.js 14 and 16. We should widen the version range to allow our projects to use it.

# Checklist:

### DIFF
--- a/plugins/mocha/package.json
+++ b/plugins/mocha/package.json
@@ -39,7 +39,7 @@
   },
   "peerDependencies": {
     "dotcom-tool-kit": "4.x",
-    "mocha": ">=6.x <=10.x"
+    "mocha": ">=6.x <=11.x"
   },
   "engines": {
     "node": "18.x || 20.x || 22.x"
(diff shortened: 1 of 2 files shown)


plugins/mocha/package.json
```diff
@@ -39,7 +39,7 @@
   },
   "peerDependencies": {
     "dotcom-tool-kit": "4.x",
-    "mocha": ">=6.x <=10.x"
+    "mocha": ">=6.x <=11.x"
```
joelcarr, line 42: **question(non-blocking):** Should we have an upper version limit?

joelcarr (approved):

plugins/mocha/package.json
rowanmanning: Yeah I considered removing it but tbh if mocha v12 changes the command line API then it's break tool kit so I think it's safer to leave it in place. This is the first mocha major version in 3 years so it's not a maintenance burden really

PR #771, opened 2025-02-04
feat!: move docker auth to a new task

rowanmanning:
# Description

We hit an annoying roadblock here. Because all the auth is done as part of the `DockerPush` task, we aren't authenticated when we try to pull a built image as part of the production deploy.

I think it's sensible to pull the auth out of the `DockerPush` task. Unfortunately this is a breaking change, but that's why we went with an `0.1.0` release originally.

# Checklist:


CyntiBinti (approved):
Nice 👌🏽

ivomurrell (approved):
**praise**: good that adding another task avoids duplication and simplifies the auth logic

PR #776, opened 2025-02-14
fix: use a new "docker" state for pushed images

rowanmanning:
# Description

We were seeing some issues with using "ci" state because it's written to in multiple jobs - we can't rely on docker images being available when we reach deploy-prod.

# Checklist:

### DIFF
--- a/plugins/hako/src/tasks/deploy.ts
+++ b/plugins/hako/src/tasks/deploy.ts
@@ -15,9 +15,8 @@ export default class HakoDeploy extends Task<{ task: typeof HakoDeploySchema }> 
   async run() {
     this.logger.info('Deploying to Hako')
     try {
-      const ciState = readState('ci')
-      const awsCredentials = ciState?.awsCredentials || {}
-      const pushedImages = ciState?.pushedImages || []
+      const awsCredentials = readState('ci')?.awsCredentials || {}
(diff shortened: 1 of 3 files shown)


CyntiBinti (approved):

plugins/hako/src/tasks/deploy.ts
```diff
@@ -15,9 +15,8 @@ export default class HakoDeploy extends Task<{ task: typeof HakoDeploySchema }>
   async run() {
     this.logger.info('Deploying to Hako')
     try {
-      const ciState = readState('ci')
-      const awsCredentials = ciState?.awsCredentials || {}
-      const pushedImages = ciState?.pushedImages || []
+      const awsCredentials = readState('ci')?.awsCredentials || {}
+      const pushedImages = readState('docker')?.pushedImages || []
```
ivomurrell, line 19: **nitpick**: (sorry it's tangential to your change but) it would be more idiomatic to use `??` here as you aren't expecting other kinds of falsy values

```suggestion
      const awsCredentials = readState('ci')?.awsCredentials ?? {}
      const pushedImages = readState('docker')?.pushedImages ?? []
```

ivomurrell (approved):
think this belies some concerning CircleCI behaviour that we'll want to investigate further but this is a suitable workaround, thanks!

PR #778, opened 2025-02-14
feat: add wrapper plugins for Docker > Hako

rowanmanning:
This adds two new wrapper plugins to build a Docker image and deploy via Hako. The reason there are two is because we want one with assets (similar to frontend-app) and one without (similar to backend app). I think I prefer being explicit with what the difference is, hence the `with-assets`.

Another option would be to introduce a _single_ wrapper plugin with an option to turn asset loading on/off. This would be neater in terms of what we have to maintain, however we'd be installing the webpack plugin for apps that don't need it.

See-Also: [CPREL-1275](https://financialtimes.atlassian.net/browse/CPREL-1275)

[CPREL-1275]: https://financialtimes.atlassian.net/browse/CPREL-1275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/containerised-app-with-assets/readme.md
```diff
@@ -0,0 +1,56 @@
+# dotcom-tool-kit/containerised-app-with-assets
+
+A bootstrap plugin that provides the minimum required Tool Kit plugins for a containerised application that's deployed to Amazon Elastic Container Service (ECS), including building and uploading assets to S3
```
ivomurrell, line 3: **nitpick**: this isn't really the minimum required set of plugins now that you're adding some batteries 😝 think it makes sense to differentiate this wrapper a bit from `containerised-app`

```suggestion
A bootstrap plugin that provides the required Tool Kit plugins for a containerised application that's deployed to Amazon Elastic Container Service (ECS,) additionally supporting building and uploading assets to S3.
```

plugins/containerised-app/readme.md
ivomurrell, line 1: **praise**: thanks for expanding on the standard plugin documentation!

plugins/containerised-app/readme.md
```diff
@@ -0,0 +1,59 @@
...
+A bootstrap plugin that provides the minimum required Tool Kit plugins for a containerised application that's deployed to Amazon Elastic Container Service (ECS).
+
+  - [`@dotcom-tool-kit/aws`](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/aws)
+  - [`@dotcom-tool-kit/cloudsmith`](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/cloudsmith)
+  - [`@dotcom-tool-kit/docker`](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/docker)
+  - [`@dotcom-tool-kit/doppler`](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/doppler)
+  - [`@dotcom-tool-kit/hako`](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/hako)
+  - [`@dotcom-tool-kit/node`](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/node)
+
+This bootstrap plugin is also preconfigured to run the `Node` task on the hook `run:local`.
```
ivomurrell, line 12: **todo**:

```suggestion
This bootstrap plugin is also preconfigured to run the `Node` task – after building via the `Webpack` task – on the command `run:local`.
```

plugins/containerised-app-with-assets/readme.md
```diff
@@ -0,0 +1,56 @@
...
+
+  - [`@dotcom-tool-kit/containerised-app`](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/containerised-app)
+  - [`@dotcom-tool-kit/upload-assets-to-s3`](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/upload-assets-to-s3)
+  - [`@dotcom-tool-kit/webpack`](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/webpack)
+
+This bootstrap plugin is also preconfigured to run the `Node` task on the hook `run:local`.
```
ivomurrell, line 9: **nitpick**: since v4, hooks are the logic to install Tool Kit config into other files, whilst commands are how we bundle tasks together

```suggestion
This bootstrap plugin is also preconfigured to run the `Node` task on the command `run:local`.
```

ivomurrell (commented):
i think the commands are sufficiently different enough to justify having two different wrappers like you've done!

CyntiBinti (approved):
woohoo 🚀

plugins/containerised-app-with-assets/schema.js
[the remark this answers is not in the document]
rowanmanning: Ah yeah we could do that 👍 will the validation work because `containerised-app` is a dependency of `containerised-app-with-assets`?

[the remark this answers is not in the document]
ivomurrell: yep, should do!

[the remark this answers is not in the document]
ivomurrell: same as how the validation works for the `docker` plugin etc. we validate the options once all the plugins have been resolved and merged, so if a schema requires an option to be given then it doesn't need to be specified in the plugin immediately depending on it, but it needs to be specified _somewhere_ in the descendants.

plugins/containerised-app/readme.md
rowanmanning: Thanks! Last minute change because I felt bad about not doing it 😂

PR #796, opened 2025-02-27
refactor: move hako deploy running into a method

rowanmanning:
# Description

We're intending on adding some more complexity to this, adding the ability to deploy an ephemeral (review) app and streaming a file into the `stdin` of this `docker run`. This will require some more async code and some error handling.

Because of this it makes sense to bundle it all into a single async function rather than only putting the `waitOnExit` into our array of promises.

Another PR will land soon to do the rest of the work and it'll be easier to review because of this 😁

# Checklist:


olusoladavid (approved):
Smooth refactor. LGTM

PR #797, opened 2025-02-27
feat!: adopt standard labels for container images

rowanmanning:
# Description

This removes a lot of the tags we've been adding to images, reducing it down to only the tag we use for deployment. It replaces many of these tags with the labels outlined in the [Standard Labels for FT Container Images proposal](https://docs.google.com/document/d/1USDbK4ptBe4VRW_3Ujv0RCgrD9JOhMHT1JjImQGvmRQ/edit).

# Checklist:

### DIFF
--- a/plugins/docker/src/image-info.ts
+++ b/plugins/docker/src/image-info.ts
@@ -6,56 +6,34 @@ export function buildImageName({ registry, name }: { registry: string; name: str
   return joinPath(registry, name)
 }
 
-// See https://docs.docker.com/reference/cli/docker/image/tag/#description
-// We're just doing some basic sanitization now to avoid git tags causing issues.
-// If the tag isn't valid then Docker will error so it's not high risk
-function sanitizeDockerTag(tag: string): string {
-  return tag.replace(/[^a-z0-9\._-]+/gi, '')
+interface ImageLabels {
+  'org.opencontainers.image.created': string
+  'org.opencontainers.image.vendor': string
+  'org.opencontainers.image.source': string | undefined
+  'org.opencontainers.image.revision': string | undefined
+  'org.opencontainers.image.version': string | undefined
+  'com.ft.system.code': string
+  'com.ft.build.url': string | undefined
+  'com.ft.source.branch': string | undefined
 }
 
-interface ImageTags {
-  deploy: string
-  gitCommit?: string
-  gitTag?: string
-  gitBranch?: string
-}
-
-function generateImageTags(ciState: CIState | null): ImageTags {
-  const tags: ImageTags = {
-    deploy: getDeployTag(ciState)
+export function generateImageLabels(systemCode: string): ImageLabels {
+  return {
+    'org.opencontainers.image.created': new Date().toISOString(),
+    'org.opencontainers.image.vendor': 'The Financial Times Ltd',
+    'org.opencontainers.image.source': process.env.CIRCLE_REPOSITORY_URL,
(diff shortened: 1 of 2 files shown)


plugins/docker/src/image-info.ts
```diff
@@ -6,56 +6,34 @@ export function buildImageName({ registry, name }: { registry: string; name: str
...
-  if (ciState?.version) {
-    tags.gitCommit = `git-${ciState.version.slice(0, 7)}`
-  }
-
-  if (ciState?.tag) {
-    tags.gitTag = `release-${ciState.tag}`
+export function generateImageLabels(systemCode: string): ImageLabels {
+  return {
+    'org.opencontainers.image.created': new Date().toISOString(),
+    'org.opencontainers.image.vendor': 'The Financial Times Ltd',
```
olusoladavid, line 23: **note:** I asked a question in the proposal document as to whether we wanted to standardize the vendor name for FT as an organization. Eg in the doc, it suggests `Financial Times`. What do you think?

plugins/docker/src/image-info.ts
```diff
@@ -6,56 +6,34 @@ export function buildImageName({ registry, name }: { registry: string; name: str
...
+    'org.opencontainers.image.source': process.env.CIRCLE_REPOSITORY_URL,
+    'org.opencontainers.image.revision': process.env.CIRCLE_SHA1
+      ? process.env.CIRCLE_SHA1.slice(0, 7)
```
olusoladavid, line 26: **curious:** Any reason why we think `CIRCLE_SHA1` may be `undefined` in CI and would it be worth flagging with an error? 

_Maybe we could also throw in some optional chaining for some one-liner_

olusoladavid (approved):
Elegant 🫡 . Just one note around vendor name and a CI curiosity, thanks

plugins/docker/src/image-info.ts
rowanmanning: Oh interesting, in the proposal it's set to `Financial Times` in the table but `The Financial Times Ltd` in the example Docker command. I'll ask for that to be corrected and I'll update this

plugins/docker/src/image-info.ts
rowanmanning: This is mostly so that we can run `DockerBuild` locally without having to manually specify all these labels. It's been useful in testing so we're being quite defensive and assuming all of the environment variables could be missing. Also this condition satisfies TypeScript where anything under `process.env` has the type `undefined | string`

plugins/docker/src/tasks/build.ts
```diff
@@ -35,13 +30,21 @@ export default class DockerBuild extends Task<{
         hookFork(this.logger, 'docker-build-create', childCreate)
         await waitOnExit('docker-build-create', childCreate)
 
+        const labels = Object.entries(generateImageLabels(imageName))
+          .filter(([, value]) => value)
+          .flatMap(([label, value]) => {
+            return ['--label', `${label}=${value}`]
```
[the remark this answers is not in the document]
ivomurrell: (it's not that Node escapes spaces, it's because you're passing the arguments directly to the spawned process rather than invoking a command via a shell. the shell (e.g., bash/zsh) is the one that automatically splits spaces into separate arguments and using `spawn` instead of `exec` means that a shell is never used – that's why you also don't need to worry about sanitisation of commands with `spawn` either!)

PR #799, opened 2025-02-28
feat: support review apps

rowanmanning:
# Description

This adds in an option to run hako deploy with an "ephemeral" app, which is a similar concept to Heroku review apps. To make this work we need to pass Hako config into the Docker command as STDIN.

This also updates the two wrapper plugins to introduce a `deploy:review` command.

This was all tested locally in cp-docker-migration-canary. It works <del>aside from an issue with name length, we're talking to Cloud Enablement about this</del> <ins>fixed</ins>

This relies on https://github.com/Financial-Times/hako-cli/pull/324

# Checklist:


plugins/containerised-app-with-assets/.toolkitrc.yml
```diff
@@ -11,7 +11,19 @@ commands:
...
-  'deploy:review': [] # Coming soon, see https://financialtimes.atlassian.net/browse/CPREL-1280
+  'deploy:review':
+    - Webpack:
+        envName: production
+    - UploadAssetsToS3
+    - DockerAuthCloudsmith
+    - DockerBuild
+    - DockerPush
+    - AwsAssumeRole:
+        roleArn: !toolkit/option '@dotcom-tool-kit/containerised-app.awsRoleArnStaging'
```
CyntiBinti, line 22:
```suggestion
        roleArn: !toolkit/option '@dotcom-tool-kit/containerised-app-with-assets.awsRoleArnStaging'
```

CyntiBinti (changes requested):
woohoo 🙌🏽  Just 1 suggestion from me!

plugins/containerised-app-with-assets/.toolkitrc.yml
rowanmanning: The `containerised-app-with-assets` plugin reuses the configurations from `containerised-app` so this is correct, see the other `!toolkit/option` lines further down this file. See [this commit](https://github.com/Financial-Times/dotcom-tool-kit/commit/4f6fac7e5a5bd93d515031d8eed5bffc3e2e5fa4) for the change and [this comment](https://github.com/Financial-Times/dotcom-tool-kit/pull/778#discussion_r1960010617) from @ivomurrell explaining why

plugins/containerised-app-with-assets/.toolkitrc.yml
```diff
@@ -11,7 +11,19 @@ commands:
...
+    - HakoDeploy:
+        asReviewApp: true
```
olusoladavid, line 24: **curious:** is there an external teardown mechanism for the review apps?

plugins/hako/src/tasks/deploy.ts
```diff
@@ -58,7 +65,29 @@ export default class HakoDeploy extends Task<{ task: typeof HakoDeploySchema }>
...
+      try {
+        const hakoConfig = await readFile(hakoConfigPath, 'utf-8')
+        child.stdin.setDefaultEncoding('utf-8')
+        child.stdin.write(hakoConfig)
+      } catch (error) {
+        child.kill('SIGTERM')
+        throw new Error(`Hako config not found at ${hakoConfigPath}`)
+      }
+    }
+    child.stdin.end()
```
olusoladavid, line 90: **praise:** beautifully done

olusoladavid (approved):
LGTM, great work. Thanks

plugins/containerised-app-with-assets/.toolkitrc.yml
CyntiBinti: Ah ah cool thanks!

plugins/containerised-app-with-assets/.toolkitrc.yml
rowanmanning: Not _yet_ but hopefully soon! For now while we're in beta we're fine with this being manual, but yeah we'll need to make sure this is in place before we migrate 80+ apps 😂

CyntiBinti (approved):
👌🏽 Lets go!

plugins/hako/src/tasks/deploy.ts
ivomurrell, line 90: **suggestion**: could avoid loading the file into memory with streams

```suggestion
      try {
        const readStream = fs.createReadStream(configPath, { encoding: 'utf8' });
        readStream.pipe(child.stdin);
      } catch (error) {
        child.kill('SIGTERM')
        throw new Error(`Hako config not found at ${hakoConfigPath}`)
      }
    }
```

plugins/hako/src/tasks/deploy.ts
```diff
@@ -58,7 +65,29 @@ export default class HakoDeploy extends Task<{ task: typeof HakoDeploySchema }>
...
+    if (this.options.asReviewApp) {
+      commandArgs.push('--ephemeral')
+      if (process.env.CIRCLE_PR_NUMBER) {
+        commandArgs.push('--pr-number', process.env.CIRCLE_PR_NUMBER)
+      }
+    }
+    const child = spawn('docker', commandArgs)
+
+    // Because we can't mount volumes in Docker images on CircleCI we have to
+    // pass the hako config via STDIN
```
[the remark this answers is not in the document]
rowanmanning: Because of [this](https://support.circleci.com/hc/en-us/articles/360007324514-How-can-I-use-Docker-volume-mounting-on-CircleCI), Docker is incapable of mounting volumes (and thus reading files) in CircleCI unless you use a machine executor which would require us to make a lot of CircleCI config changes.

[More detail here](https://github.com/Financial-Times/hako-cli/issues/321), sorry I should have linked this.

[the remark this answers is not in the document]
rowanmanning: Oh wait that wasn't what you were asking. Because the `docker run` in question is Hako, not the image _we_ built

plugins/hako/src/tasks/deploy.ts
rowanmanning: We thought about that and tried it, but the error handling becomes a pain and I think the end result was harder to understand. These config files are small (smaller than `.toolkitrc.yml`) and so we decided loading into memory wasn't a big overhead vs the added complexity.

plugins/hako/src/tasks/deploy.ts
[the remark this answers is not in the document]
ivomurrell: Sorry, I mean why can't we include the `hako-config/apps/.../app.yaml` file in the Docker image we build in the `DockerBuild` task – but the answer is because we're running the hako image here not the image we've built 😰 This approach makes sense then (though I hope we'll get to a point where Hako doesn't (officially) require Docker some day)

PR #803, opened 2025-03-03
Widen dependency range for serverless-offline peer dependency

ivomurrell:
# Description

The range now includes v13. This [version](https://github.com/dherault/serverless-offline/releases/tag/v13.0.0) drops support for Node 16 (which we've also dropped support for) and no longer will try to install `node-fetch` if fetch is not present.

This is important for us as it uses a top-level await to import `node-fetch` which causes issues when running with Node 22. Serverless will typically try to require a module then fall back to an ESM import if that fails, but with Node 22 `require`'ing a module that has any top level awaits will cause an exception to be thrown instead.

Adding v13 to the range of permissible versions will allow consumers to update their version of `serverless-offline` in order to avoid this issue as part of a Node 22 migration. Note that the later v14 is not included in the range as that requires Serverless v4 which we do not support.

# Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,10 +48,10 @@
     },
     "core/cli": {
       "name": "dotcom-tool-kit",
-      "version": "4.6.0",
+      "version": "4.7.0",
       "license": "MIT",
       "dependencies": {
-        "@dotcom-tool-kit/base": "^1.1.10",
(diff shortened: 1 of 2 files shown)


olusoladavid (approved):
Love the PR description, well explained

PR #806, opened 2025-03-07
fix: make review app URLs predictable

rowanmanning:
# Description

We now generate a hash based on the branch name and use this as a suffix on review app names. This means a couple of things:

  1. Review app names are predictable
  2. Only one review app will be created per branch, with future commits to that branch updating the existing review app instead of deploying another one

This is good for a couple of reasons:

We can now hand the review app URL off to future jobs. This is important for the n-test and cyprus plugins which expect a URL in the `review` or `staging` state.

We will stop creating a new review app for every commit, we're already seeing way too many review apps get created which is a problem while we don't have automatic deletion of old review apps.

Changes to a PR will deploy faster because most of the review app CloudFormation stack will have been created already - we'll be doing updates instead of fully creating a new stack on each commit.

[Here's an example of a review app deployed with this change](https://cp-docker-migration-canary-f81bbd.eu-west-1.ft-com-test.ftweb.tech/cp-docker-migration-canary). The suffix `f81bbd` is `example-branch` hashed. I tried running the deploy a second time and the stack was updated instead of created, **super quick**.

See-also: [CPREL-1300](https://financialtimes.atlassian.net/browse/CPREL-1300)

# Checklist:

[CPREL-1300]: https://financialtimes.atlassian.net/browse/CPREL-1300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/plugins/hako/src/tasks/deploy.ts
+++ b/plugins/hako/src/tasks/deploy.ts
@@ -1,13 +1,14 @@
 import { spawn } from 'child_process'
 import { z } from 'zod'
 import { hookFork, waitOnExit, styles } from '@dotcom-tool-kit/logger'
-import { CIState, readState } from '@dotcom-tool-kit/state'
+import { CIState, readState, writeState } from '@dotcom-tool-kit/state'
 import { Task } from '@dotcom-tool-kit/base'
 import { ToolKitError } from '@dotcom-tool-kit/error'
 import { readFile } from 'node:fs/promises'
 import { join } from 'node:path'
+import { createHash } from 'node:crypto'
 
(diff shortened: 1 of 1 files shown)


ivomurrell (commented):
looks good so far! let's see how the hako implementation goes.

ivomurrell (approved):
great! thanks for working with cloud enablement on this. we'll test this with next-static once it's released

PR #811, opened 2025-03-26
CPP-2441: Document some known issues in v4 migration

olusoladavid:
# Description
This PR documents a known issue observed during v4 migrations to help teams migrating take note of it. Specifically, `nodeVersion` rename to `cimgNodeVersions` is not being flagged which means CI configs can regress to the default Node 18 from Node 22.

Also, adds an early pointer to troubleshooting help note when the user runs the command in Step 0 and runs into npm peer-deps errors. This is useful because many devs will switch context from the docs to a terminal and might be tempted to go down a rabbit hole to debug the error instead of just reading the troubleshooting note after the command

# Checklist:


docs/migration-guides/v4.md
```diff
@@ -475,4 +475,7 @@ You'll need to [talk to the Platforms team](https://financialtimes.enterprise.sl
 
 Once `npx dotcom-tool-kit --help` is no longer reporting any configuration or plugin errors, run `npx dotcom-tool-kit --install`. This will regenerate your `package.json` and `.circleci/config.yml`. If you've previously opted out of having your `.circleci/config.yml`, the new version of Tool Kit might not be able to validate it; [get in touch with the Platforms team](https://financialtimes.enterprise.slack.com/archives/C02TRE2V2Q1) and we'll help you update it. Note that Tool Kit v4 makes CircleCI config file generation much more configurable and flexible, so in many cases you'll no longer have to opt out of config file management.
 
+> [!NOTE]
+> The `nodeVersion` option for the CircleCI plugin has been renamed to `cimgNodeVersions` which takes an array value but our `--help` validator doesn't currently report that. Please double-check the Node version generated in your CircleCI config to ensure it is the same version specified in the `.toolkitrc.yml` file and has not reverted to a default `v18.x`. We are working on a fix to ensure the validator reports this when you run `npx dotcom-tool-kit --help`.
```
ivomurrell, line 479: **todo**: `--help` doesn't run any special validation – options are validated on any Tool Kit invocation. we only suggest using `--help` in this guide as it's a non-destructive call. let's update this wording so we don't inadvertently imply otherwise.

```suggestion
> The `nodeVersion` option for the CircleCI plugin has been renamed to `cimgNodeVersions` which takes an array value but our option validation doesn't currently report that. Please double-check the Node version generated in your CircleCI config to ensure it is the same version specified in the `.toolkitrc.yml` file and has not reverted to a default `v18.x`. We are working on a fix to ensure the validator reports this when you run a Tool Kit command.
```

olusoladavid: Done, thanks

ivomurrell (approved):
thanks for keeping these up to date!

PR #813, opened 2025-03-31
feat: add a node-test plugin

rowanmanning:
# Description

This adds a node-test plugin using `node:test#run()`. This plugin only works in Node.js 20 and 22, with no support for Node.js 18. It's possible to add support for Node.js 18.17.0+ however it complicates the plugin a lot - I don't think it's worth doing seeing as 18 goes end-of-life in April.

The options for this task go against some of the Tool Kit design principals, mostly the fact that we rely on task options instead of a central config file. This is because `node --test` is configured completely via CLI or JS options and there's no config file.

We also make a compromise to support both Node.js 20 _and_ 22. Node.js 22 supports glob-patterns out of the box however Node.js 20 does not. This means we need to glob for files ourselves. I'm using the same patterns as Node.js 22's defaults so we can just remove these once we drop support for Node.js 20.

This supersedes #688.

# Checklist:

- [x] My commit messages are [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)


plugins/node-test/src/tasks/node-test.ts
```diff
@@ -0,0 +1,87 @@
...
+  '**/test/**/*.?(c|m)js'
+]
+
+// We don't want to run tests against files under "node_modules"
+const defaultIgnorePatterns = ['**/node_modules/**']
+
+const NodeTestSchema = z
+  .object({
+    concurrency: z
+      .number()
```
ivomurrell, line 30: **suggestion**: `concurrency` should also allow a `boolean`:

> If `true`, it would run `os.availableParallelism() - 1` test files in parallel.

https://nodejs.org/docs/latest-v20.x/api/test.html#runoptions

we might want to reuse the [logic](https://github.com/Financial-Times/dotcom-tool-kit/blob/b668d28d931ac7aa1620f2155f0d503507377b57/plugins/jest/src/tasks/jest.ts#L16-L30) we have in the jest plugin that does a better approximation for `availableParallelism` than Node natively provides when running on a shared CircleCI server. whilst checking this is all still true i've noticed that this logic no longer works since our docker executors were [upgraded](https://discuss.circleci.com/t/docker-executor-infrastructure-upgrade/52282) 😅 i'll make a ticket to fix that (it's because we were relying on the old cgroupsv1 fs api to guess the CPUs allocated to us)

plugins/node-test/src/tasks/node-test.ts
```diff
@@ -0,0 +1,87 @@
...
+      .describe('Glob patterns for files to ignore when running tests')
+  })
+  .describe('Run tests via the Node.js built-in test runner')
+
+export { NodeTestSchema as schema }
+
+export default class NodeTest extends Task<{ task: typeof NodeTestSchema }> {
+  async run() {
+    try {
+      const cwd = process.cwd()
```
ivomurrell, line 55: **todo**: the cwd now gets passed in to the task in a context object – this allows packages within a monorepo to have different working directories during the same Tool Kit invocation

```suggestion
  async run({ cwd }) {
    try {
```

plugins/node-test/src/tasks/node-test.ts
```diff
@@ -0,0 +1,87 @@
...
+      const files = await glob(this.options.files, { cwd, ignore: this.options.ignore })
+
+      return new Promise<void>((resolve, reject) => {
+        let success = true
+        run({
```
ivomurrell, line 60: **suggestion**: I don't think we can ever hope to be exhaustive with the options we expose in the schema so I'd suggest we add an escape hatch `customOptions` field with an `unknown` type that we can merge with. though i'd also suggest a `watch` job as another option to explicitly add 😄

plugins/node-test/src/tasks/node-test.ts
```diff
@@ -0,0 +1,87 @@
...
+          forceExit: this.options.forceExit
+        })
+          .on('test:fail', () => {
+            success = false
+          })
+          .on('end', () => {
+            if (success) {
+              return resolve()
+            }
+            reject(new ToolKitError('Some tests did not pass'))
```
ivomurrell, line 72: **question**: does it print out more helpful error messages to stdout?

plugins/node-test/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": "../../tsconfig.settings.json",
+  "compilerOptions": {
+    "outDir": "lib",
+    "rootDir": "src",
+    "types": ["glob", "node"]
```
ivomurrell, line 6: **todo**: `glob` is not used *glob*ally so we don't need to add it to the *glob*al scope
```suggestion
```

plugins/node-test/src/tasks/node-test.ts
```diff
@@ -0,0 +1,87 @@
...
+        let success = true
+        run({
+          concurrency: this.options.concurrency || false,
+          files,
+          forceExit: this.options.forceExit
+        })
+          .on('test:fail', () => {
+            success = false
+          })
+          .on('end', () => {
```
ivomurrell, line 68: **thought**: i find it weird that i can't find any mention of what `'end'` means exactly and what is even readable from this stream in the docs

plugins/node-test/src/tasks/node-test.ts
ivomurrell, line 72: **praise**: i like this approach to making sure we exit with an error if a test fails! (though also weird that there isn't a builtin promise API?)

plugins/node-test/src/tasks/node-test.ts
```diff
@@ -0,0 +1,87 @@
...
+import { run } from 'node:test'
+// Ignored because `node:test/reporters` is available in Node.js 20 but the root ESLint config
+// supports Node.js 18+. It seems overkill to add an overriding ESLint config
+// eslint-disable-next-line import/no-unresolved
+import { spec } from 'node:test/reporters'
+import { Task } from '@dotcom-tool-kit/base'
+import { ToolKitError } from '@dotcom-tool-kit/error'
+import { z } from 'zod'
+
+// We need to maintain this list of file patterns if we want to make this plugin work consistently
```
ivomurrell, line 11: **praise**: the compatibility is very well considered

ivomurrell (changes requested):
this will be a great addition to our lineup of plugins (once a few adjustments have been made,) thanks!

plugins/node-test/src/tasks/node-test.ts
```diff
@@ -0,0 +1,87 @@
...
+          .on('test:fail', () => {
+            success = false
+          })
+          .on('end', () => {
+            if (success) {
+              return resolve()
+            }
+            reject(new ToolKitError('Some tests did not pass'))
+          })
+          .compose(new spec())
```
ivomurrell, line 74: **thought**: maybe we could configure the tests to also stream into a file in a jUnit format so that CircleCI will report our tests nicely? though we currently [assume](https://github.com/Financial-Times/dotcom-tool-kit/blob/b668d28d931ac7aa1620f2155f0d503507377b57/orb/src/jobs/test.yml#L14) jUnit files get put into a Jest directory so i'd say we should work on this separately.

plugins/node-test/src/tasks/node-test.ts
rowanmanning: It's a `TestStream` which has [_some_ docs](https://nodejs.org/api/test.html#class-testsstream) but mostly it inherits from [`ReadableStream`](https://nodejs.org/api/stream.html#class-streamreadable) which has an `end` event that gets called when the stream is closed.

plugins/node-test/src/tasks/node-test.ts
rowanmanning: Yep it outputs everything that the [`spec` reporter does](https://nodejs.org/api/test.html#test-reporters). Other reporters are available but this one has a nice balance of including useful error/pass details but without _flooding_ the output. We could make it configurable in future if we want to 👍

rowanmanning: Yeah I think the golden path is "use the `node --test` CLI" and they're only expecting this API to be used by other testing libraries that layer over the top. No promises, only streams 👍

plugins/node-test/src/tasks/node-test.ts
rowanmanning: Nice, TIL

plugins/node-test/src/tasks/node-test.ts
rowanmanning: I'll allow it to be a boolean and link to the docs 👍

plugins/node-test/src/tasks/node-test.ts
rowanmanning: I like this idea but I think for now I'd rather add options as we need them to avoid potential breaking changes if we want to tweak options in the future. There are also a bunch of new options between Node.js 20 and 22.

Agree with adding `watch`, I'll do that

rowanmanning: Sorry that wasn't clear, I mean with `customOptions` I assume we'll merge these with the explicit options we define. So something like:

```js
run(Object.assign({}, customOptions, { files, watch, etc... }));
```

Then let's say we want to formally define the `setup` option in future - would that minor version of Tool Kit potentially mess with whatever someone threw into `setup` in `customOptions`? Maybe I'm just being overly cautious on this or misunderstanding what you're suggesting.

ivomurrell: I think that's a good point and a good reason to do
```ts
run(Object.assign({}, { files, watch, etc... }, customOptions));
```
instead. Let's trust the consumer not to break things!

rowanmanning: Is `customOptions` a fully new option for the plugin then? So I'd specify:

```yml
  test:local:
    - NodeTest:
        watch: true # or whatever
        customOptions:
          example: 'hello'
```

And we only assert that it's `Record<string, unknown>` in Zod?

ivomurrell: Yep!

rowanmanning: Done!

plugins/node-test/src/tasks/node-test.ts
```diff
@@ -0,0 +1,98 @@
...
+          .pipe(process.stdout)
+      })
```
ivomurrell, line 87: **nitpick**: `readable.compose` is an experimental function which will likely be unfamiliar to other devs. i think it would be more readable (and simpler seeing as we do want a 'closed circuit' here rather than a composable stream) to use `stream.pipeline` to pipe each of these streams. this comes with the additional benefit that we _should_ be able to use a promise API to `await` on the stream closing, as by default the `end` option will be set which implies that it will 'End the destination stream when the source stream ends.'

```suggestion
      let success = true
      await pipeline(
        run(Object.assign({ concurrency, files, forceExit, watch }, customOptions)).on('test:fail', () => {
          success = false
        }),
        new spec(),
        this.logger.child({ process: 'node-test' })
      )
      if (!success) {
        throw new ToolKitError('Some tests did not pass')
      }
```

plugins/node-test/src/tasks/node-test.ts
```diff
@@ -0,0 +1,98 @@
...
+        let success = true
+        run(Object.assign({ concurrency, files, forceExit, watch }, customOptions))
+          .on('test:fail', () => {
+            success = false
+          })
+          .on('end', () => {
+            if (success) {
+              return resolve()
+            }
+            reject(new ToolKitError('Some tests did not pass'))
```
ivomurrell, line 83: **suggestion**: in our other plugins where we print out loads of logs we [include](https://github.com/Financial-Times/dotcom-tool-kit/blob/b668d28d931ac7aa1620f2155f0d503507377b57/lib/logger/src/helpers.ts#L156) a little hint to read the previous logs for test failure reasons in this error message

```suggestion
            reject(new ToolKitError('some tests did not pass, error output has been logged above ☝️'))
```

plugins/node-test/src/tasks/node-test.ts
```diff
@@ -0,0 +1,98 @@
...
+          })
+          .compose(new spec())
+          .pipe(process.stdout)
```
ivomurrell, line 86: **issue**: piping directly to stdout will mean that our logger hooks will be skipped and none of the logs will be prefixed with which Tool Kit task they're coming from. usually we use [`hookFork` or `hookConsole`](https://github.com/Financial-Times/dotcom-tool-kit/blob/b668d28d931ac7aa1620f2155f0d503507377b57/lib/logger/src/helpers.ts) to make sure there's still that context but in this case we can pipe to our logger. if we add a `process` metadata field that will be included in the logs too.

```suggestion
          .pipe(this.logger.child({ process: 'node-test' }))
```

plugins/node-test/src/tasks/node-test.ts
rowanmanning: Again, looking into this but I get errors. The first is:

```
 ×  AggregateError: 1 error running tasks for test:local
╰─ ×  NodeTest → ToolKitError: node test failed to run
   i  The "streams[stream.length - 1]" property must be of type function. Received an instance of DerivedLogger
```

If I switch `this.logger.child` back to `process.stdout` then the error just moves up to the line above:

```
 ×  AggregateError: 1 error running tasks for test:local
╰─ ×  NodeTest → ToolKitError: node test failed to run
   i  The "streams[stream.length - 1]" property must be of type function. Received an instance of WriteStream
```

I remember reading something about why `pipeline` isn't used in the docs for this, will see if I can find it.

plugins/node-test/src/tasks/node-test.ts
ivomurrell: oops yeah we'll need to do a similar transformation as we do in the [`hookFork` function](https://github.com/Financial-Times/dotcom-tool-kit/blob/b668d28d931ac7aa1620f2155f0d503507377b57/lib/logger/src/helpers.ts#L109-L122) in order for the piping to work. i can do that separately if you prefer (by adding a new `hookReadable` helper function which extracts that logic to the `logger` lib) and we can pipe to `stdout` for now

plugins/node-test/src/tasks/node-test.ts
ivomurrell: my code was importing `pipeline` from [`node:stream/promises`](https://nodejs.org/docs/latest-v18.x/api/stream.html#streams-promises-api) _not_ `node:stream`, so it doesn't require a callback as the final argument. sorry, should have been clearer!

rowanmanning: I probably would have got there eventually if I wasn't rushing to standup 😂 thanks!

plugins/node-test/src/tasks/node-test.ts
rowanmanning: Would you rather do that? Feels like maybe it'd be best to so we don't add a bunch of non-test-related code to this plugin.

ivomurrell: well we definitely don't want it in this plugin, don't feel strongly about whether the new logger additions go in this PR or separate

ivomurrell (approved):
looks great, thanks for fielding all my nitpicks!

PR #823, opened 2025-04-08
Add Dependabot config to group updates

ivomurrell:
# Description

We never enabled Dependabot/the auto merger in this repository! Let's get it configured.

# Checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,27 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    commit-message:
+      prefix: 'fix:'
+      prefix-development: 'chore:'
+    groups:
+      aws-sdk:
+        patterns:
+          - '@aws-sdk/*'
+        update-types:
+          - 'minor'
+          - 'patch'
+      development-dependencies:
+        dependency-type: 'development'
+        update-types:
+          - 'minor'
+          - 'patch'
+      reliability-kit:
+        patterns:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,57 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    commit-message:
+      prefix: 'fix:'
```
olusoladavid, line 8: **question:** why do we use `fix:` prefix for non-dev dependencies and not `chore:` ?

.github/dependabot.yml
```diff
@@ -0,0 +1,57 @@
...
+          - '@aws-sdk/*'
+        update-types:
+          - 'minor'
+          - 'patch'
+      development-dependencies:
+        dependency-type: 'development'
+        update-types:
+          - 'minor'
+          - 'patch'
+      origami:
```
olusoladavid, line 22: **q:** is it worth trimming the fat in this dependabot config and only include dependency groups that we actually use? for example, from first glance, Origami, page-kit, privacy, x-dash look oop

olusoladavid (approved):
Thanks for setting this up! Useful automation

.github/dependabot.yml
ivomurrell: Because we'll want to create a new release when we update those and release-please will only create releases for fix and feat commits

olusoladavid: I see. Makes sense to me. Thanks

PR #882, opened 2025-04-17
Delete unused dependencies from packages

ivomurrell:
# Description

Dependabot has created PRs for all of these dependencies and thus highlighted how they aren't actually used anywhere. `hook-std` was initially used when prototyping hooking logs into stdout, we've removed `js-yaml` already so don't need `@types/js-yaml` and `chai` seems to have just been added accidentally.

# Checklist:


olusoladavid (approved):
Thanks for the cleanup

PR #883, opened 2025-04-17
Bump to new major version of next-router package

ivomurrell:
# Description

All the new major version does is drop support for Node 18 and Node 20. I think dropping support for Node 20 is a mistake that will be remedied in due course. Whilst Tool Kit hasn't dropped support for Node 18 yet we're right on the precipice of doing so (Node 18 is EOL at the end of the month) and besides next-router isn't using any non-Node 18 code yet. Keeping this dependency up to date is more important because changes are regularly being made to next-router and teams will want to keep their local versions in sync with the deployed service.

# Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6204,12 +6204,21 @@
       }
     },
     "node_modules/@dotcom-reliability-kit/crash-handler": {
-      "version": "3.0.2",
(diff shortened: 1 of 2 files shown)


olusoladavid (approved):
Pragmatic!

PR #923, opened 2025-05-08
fix: don't try to repeat a string a non-integer number of times

apaleslimghost:
lol and indeed lmao

https://financialtimes.enterprise.slack.com/archives/C3TJ6KXEU/p1746713725780519


Shamarni96 (approved):
😆 Thanks!

ivomurrell:
closing as it turned out the issue was with the terminal reporting a negative number of columns rather than an odd number. the change in this PR is redundant as `String.prototype.repeat` will [already](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.repeat) truncate decimals passed to it.

PR #927, opened 2025-05-09
Use correct import for `tar` module

ivomurrell:
# Description

It needs to be a star import instead of a default import. The `tar` package is a TypeScript project which [doesn't](https://github.com/isaacs/node-tar/blob/d3cd4392d4a35a8f1e10287a07e57da6ae079e78/src/index.ts) have a default export. Because it's TypeScript it will set `__esModule` to true to tell other code generated by TypeScript that it doesn't need to wrap the whole package in a `{ default: x }` object. However, for some reason, TypeScript doesn't then warn us that there isn't a default object and happily lets us incorrectly import and call the module as if there is 🤷 Not sure why.

Fixes the task erroring as [reported](https://financialtimes.slack.com/archives/C3TJ6KXEU/p1746794015762799) by @rowanmanning.

# Checklist:


rowanmanning (approved):
At least it was a nice easy fix!

olusoladavid (approved):
Thanks for digging into this

PR #935, opened 2025-05-16
Fix error for using old `nodeVersion` option in `circleci` plugin not throwing

ivomurrell:
# Description

We need to set the `passthrough` field so that zod doesn't strip unrecognised keys before passing the options to the refinement function. This is what we did for other option deprecation errors but was missed with this one. 

# Checklist:


olusoladavid (approved):
Big win! Thank you for figuring it out

PR #936, opened 2025-05-16
CPP-2497 Support setting custom Hako environments in `containerised-app` plugins

### LINKED ISSUE #892: Support configurable environments in containerised-app plugins
Hiya.

At the moment the 2 containerised-app plugins will deploy to the `ft-com-test-eu` and `ft-com-prod-(eu|us)` Hako environments ([example](https://github.com/Financial-Times/dotcom-tool-kit/blob/a6a349beb39c68cadf1c13a78216085aa541d17e/plugins/containerised-app/.toolkitrc.yml#L29-L31)).

As we add new environments for different use cases we'll need this list to be configurable - possibly with the existing ones as defaults?

ivomurrell:
# Description

This allows flexibility for when apps want to deploy to custom environments.

We drop the old `multiregion` option as part of this change as it can now be better expressed by setting the `hakoProductionEnvironments` option, rather than carving out a special case for the ft-com environments. We could have preserved backwards-compatibility with this change – or even continued supporting the option – but it feels better to rip the bandage off now whilst not many apps have it set and to encourage one way of defining multiple environments.

Closes #892.

# Checklist:


rowanmanning (approved):
The stuff in the containerised app plugins looks good to me and this is a great improvement. The rest of the stuff makes sense but I'm not super familiar with these internals so you may also want Kara to take a look?

core/cli/package.json
```diff
@@ -21,6 +21,7 @@
   },
   "devDependencies": {
     "@jest/globals": "^29.7.0",
+    "@relmify/jest-serializer-strip-ansi": "^1.0.2",
```
olusoladavid, line 24: **non-blocking:** should we use an exact pinned version of this package? Given that it has not been maintained in a while and only has one author. Just to reduce our supply chain attack surface (thankfully it has no deps at the moment)

core/cli/test/options.test.ts
```diff
@@ -5,13 +5,16 @@ import * as fs from 'node:fs/promises'
 import type { Valid } from '@dotcom-tool-kit/validated'
```
olusoladavid, line 5: **thought:** should we also add a test for substituting non-string values?

plugins/containerised-app-with-assets/.toolkitrc.yml
```diff
@@ -34,21 +33,12 @@ commands:
...
-    # property under the !toolkit/if-defined if multiregion is true. We'll refactor this later when
-    # we decide what new YAML tags we need
     - HakoDeploy:
-        environments:
-          - ft-com-prod-eu
-        !toolkit/if-defined '@dotcom-tool-kit/containerised-app.multiregion':
-          environments:
-            - ft-com-prod-eu
-            - ft-com-prod-us
+        environments: !toolkit/option '@dotcom-tool-kit/containerised-app.hakoProductionEnvironments'
```
olusoladavid, line 42: **question:** Do we need to have a discussion around how and where we store this environment config? It's currently [hardcoded in the Hako plugin](https://github.com/Financial-Times/dotcom-tool-kit/blob/fc6d98da4c237557a9e7c63cce2fc0f4e205f173/plugins/hako/src/tasks/deploy.ts#L13) but since we are now allowing free user input, I think it would make sense to have a place to store them independently of plugin code changes. Also, I expect eng users would naturally not know what values are supported and from what I can see atm, we are not doing much error-handling in the Hako deploy task for bad user input

olusoladavid (commented):
Great and thoughtful implementation! Added some thoughts. Thank you

core/cli/package.json
ivomurrell: I don't think that's necessary for a dev dependency. If a security bug is found in the package presumably it would be easier to update if the version isn't pinned?

plugins/containerised-app-with-assets/.toolkitrc.yml
ivomurrell: aha this might be tangential to the original question but this highlighted to me that we'll still get errors if we pass alternative environments to the `containerised-app` plugin as the `HakoDeploy` task will fail to validate them. i'll update the `HakoDeploy` schema in a separate revision to allow any string (or something similar that will still allow the rest of the mapping logic to work.)

back to the original point, what kind of environment config are you thinking of here, Shola?

core/cli/package.json
olusoladavid: In a day and age where credentials and keys hang around on developers' machines, I'd argue that dev dependencies are just as potent an attack surface as any other dependency. The attack vector I'm looking at for this case is a situation where an attacker gets control of this package (eg compromised maintainer), and publishes a new patch/minor version of the package with malicious code in it (eg stealing creds on the machine) that executes when tests run. Either through `dependabot` or basic housekeeping, we pick this change up, mindlessly apply it (or maybe even look at the legit-looking changelog) as it is not a breaking change. This is actually a common attack vector for npm packages. See [here ](https://therecord.media/malware-found-in-npm-package-with-millions-of-weekly-downloads)and [here](https://www.blazeinfosec.com/post/dependency-confusion-exploitation).

My recommendation to pin the version is quite specific to this package. A quick inspection shows that it is a very small module with a jest serializer interface and has [no external dependencies.](https://github.com/relmify/jest-serializer-strip-ansi/blob/main/package.json). This means the code is essentially in a _"complete"_ state. Only possible change I can see is: we decide to migrate to a major version of `jest` that introduces a breaking change to the serializer interface - in which case, this package will stop working for us and would either need the maintainer to update it with a corresponding breaking change or we find a workaround. Pinning the version however mitigates the risk of a pulling in a malicious update in-between while remaining intentional about necessary update.

plugins/containerised-app-with-assets/.toolkitrc.yml
olusoladavid: @ivomurrell - my thoughts here were: since we are beginning to standardise (or at least set a convention) of storing hako configs in repos like https://github.com/Financial-Times/cp-hako-config , would it be possible for us to pull the environment mappings and values we need from that repo into the Hako plugin as a dependency and single source of truth?
cc: @rowanmanning

ivomurrell: This could be a solution in the future once our environments have been solidified and we don't expect many further changes, though I don't think we're at that point yet as we're currently envisioning creating new environments for apps like next-router and preflight. My proposed solution (which I'll push to this branch once I've got it properly working) is to assume the continuation of the current naming standard of ending the environment name with something like `-prod-eu` or `-test-us` and infer the mappings from that. We can reevaluate in the future once environments are no longer in flux or if we run into an environment name that doesn't follow that assumption.

core/cli/package.json
ivomurrell: Thanks for this writeup, I think this all makes sense. I've pinned the dependency like suggested. It would be interesting to see what other dependencies would benefit from pinning. I'm also interested in seeing how Dependabot handles updates to pinned dependencies.

rowanmanning (approved):
Nice 👍

olusoladavid (approved):
Top work

PR #941, opened 2025-05-20
Allow SSH agent forwarding to a Docker build container

ivomurrell:
# Description

This is particularly useful if you have npm dependencies on private repositories, [as is the case for the next-retention project](https://financialtimes.slack.com/archives/C085773KPRP/p1747398617908099). Let's disable it by default though as most projects won't need it and we can avoid passing around security credentials unnecessarily.

The option is documented [here](https://docs.docker.com/reference/cli/docker/buildx/build/#ssh).

# Checklist:


olusoladavid (approved):
Lgtm, thanks

PR #961, opened 2025-05-28
CPP-2423 Add support for custom Hako ephemeral apps (such as canaries)

ivomurrell:
# Description

Cloud Enablement have suggested we use ephemeral apps to support canary releases. This would allow us to use either the same Hako config or a custom manifest (which might be useful if we want to reduce resource allocation) and still deploy to the same environment. To facilitate this, we need to add support for custom ephemeral app IDs to the `hako` plugin so that we can have a stable ID to 1. limit the number of ephemeral canaries deployed to just one unique instance, and 2. to ensure we have a predictable URL we can point to from CDN config. It might be a little weird that the canaries are deleted after a period of inactivity, same as all other ephemeral apps, but we already tend to ping ops if canaries are left lingering for more than a few hours so it's good to make sure they will eventually be cleaned up tbh!

Additionally I've added a `HakoDelete` task which will delete an ephemeral app. This will allow us to scale down canaries once we've confirmed a release is okay and migrate all traffic to a production instance.

I'm still avoiding adding builtin support for canary apps as only a limited number of projects require them and it should be fairly trivial to get them working in custom Tool Kit configuration. I intend to use these new features as part of the next-router migration as that system uses canary releases currently.

# Checklist:


rowanmanning (commented):
Few comments that are more nitpicky/questions

plugins/hako/src/tasks/deploy.ts
```diff
@@ -86,32 +73,41 @@ export default class HakoDeploy extends Task<{ task: typeof HakoDeploySchema }>
...
-      commandArgs.push('--ephemeral', '--ephemeral-id', hash)
-      writeState('review', { url: `https://${name}-${hash}.${awsRegion}.${domain}` })
+      ephemeralId = hash
+      writeState('review', { url: `https://${name}-${ephemeralId}.${awsRegion}.${domain}` })
+    }
+    if (ephemeralId) {
+      commandArgs.push('--ephemeral', '--ephemeral-id', ephemeralId)
     } else {
       writeState('staging', { url: `https://${name}.${awsRegion}.${domain}` })
     }
```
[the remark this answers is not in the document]
ivomurrell: yes, both of those are intentional, though not _too_ much thought was put into them and this handling of slightly conflicting states was sort of arbitrarily chosen. let's try and put a bit more thought into them now! the choices were largely informed by assuming you want to use custom ephemeral IDs for creating canaries but maybe there are other use cases we should consider?

`asReviewApp` and `customEphemeralId` were envisioned as mutually exclusive as semantically it doesn't make sense to have a review canary app. but also practically it's largely redundant to set them both – `asReviewApp` largely serves as creating unique ephemeral apps for a given branch which it won't be able to do if a `customEphemeralId` overrides it, and the difference between writing to the `review` state and to the `staging` state is paper-thin for hako apps at the moment. so maybe the current behaviour is justified and i should add a clarification to one of the descriptions that one overrides the other?

not setting either of the states when `customEphemeralId` is set is also informed by our current canary workflows, which don't currently do end to end testing afaict. i avoided setting the states to avoid the canary clobbering other deployments setting the state. i think this is fine for now as we aren't using these states in anger in the `hako` plugin (`HakoDelete` reuses the logic to construct a predictable app name rather than relying on state.) not sure it qualifies for documenting as idk if anyone would expect it to do otherwise.

plugins/hako/src/tasks/deploy.ts
```diff
@@ -86,32 +73,41 @@ export default class HakoDeploy extends Task<{ task: typeof HakoDeploySchema }>
       environment.name
     ]
     const domain = hakoDomains[environment.stage]
+    let ephemeralId = this.options.customEphemeralId
```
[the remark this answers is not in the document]
ivomurrell: I've tried to refactor it so the control flow and order of precedence for `ephemeralId` is clearer – what do you think?

plugins/hako/readme.md
```diff
@@ -23,10 +23,25 @@ plugins:
...
-| Property                | Description                                                       | Type            | Default |
-| :---------------------- | :---------------------------------------------------------------- | :-------------- | :------ |
-| `asReviewApp`           | whether to deploy as a temporary review app, used for code review | `boolean`       | `false` |
-| **`environments`** (\*) | the Hako environments to deploy an image to                       | `Array<string>` |         |
+| Property                  | Description                                                                                                                                                                 | Type            | Default |
+| :------------------------ | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :-------------- | :------ |
+| `asReviewApp`             | whether to deploy as a temporary review app, used for code review                                                                                                           | `boolean`       | `false` |
```
olusoladavid, line 28: **thought:** What do we think about refactoring this task into different strategies with common bits extracted into lib functions/methods. I think the logical complexity we have identified here is that we are beginning to overload this task and now it is doing too many things. For example, I can see 3 strategies already: `HakoDeployReviewApp`, `HakoDeployCanary`, `HakoDeployProd`.

As a mild suggestion, one approach might be to make them into separate tasks like the `HakoDelete` one (which looks really clean), another approach might be to introduce a `deploymentTarget` or `stage` option with values: `'canary' | 'review' | 'prod' | 'staging' etc` as use it to internally determine what strategy to apply or make the conditional branching cleaner. Either option would effectively deprecate the `asReviewApp` option for example, which appears might need to evolve as it is.

plugins/hako/src/tasks/deploy.ts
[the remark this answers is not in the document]
rowanmanning: I'm not Emilio but I think it's clearer now 👍

plugins/hako/readme.md
rowanmanning: I don't hate rethinking the `asReviewApp` option. If we make a breaking change to the `hako` plugin then we can absorb it into a non-breaking change to the containerised-app plugins.

ivomurrell: One of the reasons we added task options was so that we could merge similar tasks into one task which can be configured at both the plugin and command level – we previously had a lot of tasks like `TypeScriptWatch` and `JestCi` that made slight tweaks to a shared task logic. This contributed to both an increase in boilerplate and coupling the tasks with specific use cases and commands rather than allowing flexibility.

With that context out of the way, I don't think the logic here diverges enough to justify splitting out into a different task and I don't agree that it's getting overloaded. I think this can be seen by how small the sole conditional block for `asReviewApp` currently is. Excluding some validation it's two lines which modify existing options and behaviour rather than bolting on something new.

I like the idea of an internal enum for the detected deployment stage so that the logic is clearer, though I don't think the code is complex enough to warrant it right now.

plugins/hako/readme.md
```diff
@@ -23,10 +23,25 @@ plugins:
...
-| `asReviewApp`           | whether to deploy as a temporary review app, used for code review | `boolean`       | `false` |
-| **`environments`** (\*) | the Hako environments to deploy an image to                       | `Array<string>` |         |
+| Property                  | Description                                                                                                                                                                 | Type            | Default |
+| :------------------------ | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :-------------- | :------ |
+| `asReviewApp`             | whether to deploy as a temporary review app, used for code review. overrides the `customEphemeralId` option with its own hash of the git branch.                            | `boolean`       | `false` |
+| `customEphemeralId`       | ID that is used by Hako to identify a particular ephemeral app                                                                                                              | `string`        |         |
```
olusoladavid, line 29: question(non-blocking): should we just call this `ephemeralId`? and mention that the default is a hash when not provided? Also, keeps things consistent with say `HakoDelete` options _(except I'm misundertsanding)_

similarly, just `ephemeralManifest` as opposed to `customEphemeralManifest`?

olusoladavid (approved):
🥇

PR #964, opened 2025-05-29
fix: ensure tasks exit the command on failure

rowanmanning:
# Description

Many of the tasks added by CP Reliability rethrow errors to ensure that we catch and explain errors that are outside of our control, e.g. loading a file that doesn't exist. We wanted these to be Tool Kit errors so that we get nice error rendering. We also wanted to just throw normal errors for things like validation errors or missing data.

The issue is that we didn't know about the `exitCode` on Tool Kit error. This means that any `Error` instances that we wrap will not exit the command early. This is most notable in the Docker build/deploy commands: If the Docker build fails then we still move on to try and deploy even though there's no point. This results in longer build times, unnecessary tasks being run, and confusing error messages.

We now no longer wrap Tool Kit errors themselves, this allows us to throw them with the process exit code intact.

# Review notes

There's a little bit of whitespace churn due to indentation changes, [easier to review with whitespace changes ignored](https://github.com/Financial-Times/dotcom-tool-kit/pull/964/files?w=1).

# Checklist:

### DIFF
--- a/plugins/aws/src/tasks/assume-role.ts
+++ b/plugins/aws/src/tasks/assume-role.ts
@@ -43,13 +43,13 @@ export default class AwsAssumeRole extends Task<{ task: typeof AwsAssumeRoleSche
       writeState('ci', { awsCredentials })
       this.logger.info(`Saved AWS credentials to "ci" state with session name "${RoleSessionName}"`)
     } catch (err) {
-      if (err instanceof Error) {
-        const error = new ToolKitError('failed to assume AWS role')
-        error.details = err.message
+      // We wrap non-ToolKitError errors to ensure that they exit the process
+      if (err instanceof Error && !(err instanceof ToolKitError)) {
+        const error = new ToolKitError(err.message)
+        error.exitCode = 1
         throw error
-      } else {
-        throw err
       }
+      throw err
     }
   }
 }
--- a/plugins/node-test/src/tasks/node-test.ts
+++ b/plugins/node-test/src/tasks/node-test.ts
@@ -107,33 +107,23 @@ export { NodeTestSchema as schema }
 
 export default class NodeTest extends Task<{ task: typeof NodeTestSchema }> {
   async run({ cwd }: TaskRunContext) {
-    try {
-      const { customOptions, files: filePatterns, forceExit, ignore, watch } = this.options
-      let { concurrency } = this.options
-      // the default concurrency limit needs to be corrected in CircleCI
-      if (concurrency === true && process.env.CIRCLECI) {
-        concurrency = (await guessCircleCiThreads()) - 1
-      }
-      const files = await glob(filePatterns, { cwd, ignore })
+    const { customOptions, files: filePatterns, forceExit, ignore, watch } = this.options
+    let { concurrency } = this.options
+    // the default concurrency limit needs to be corrected in CircleCI
+    if (concurrency === true && process.env.CIRCLECI) {
+      concurrency = (await guessCircleCiThreads()) - 1
+    }
+    const files = await glob(filePatterns, { cwd, ignore })
 
-      let success = true
-      const testStream = run(Object.assign({ concurrency, files, forceExit, watch }, customOptions))
-      testStream.on('test:fail', () => {
-        success = false
-      })
+    let success = true
+    const testStream = run(Object.assign({ concurrency, files, forceExit, watch }, customOptions))
+    testStream.on('test:fail', () => {
+      success = false
+    })
 
-      await pipeline(testStream, new spec(), createWritableLogger(this.logger, 'node-test'))
-      if (!success) {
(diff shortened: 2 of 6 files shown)


plugins/node-test/src/tasks/node-test.ts
```diff
@@ -130,6 +130,7 @@ export default class NodeTest extends Task<{ task: typeof NodeTestSchema }> {
       if (err instanceof Error) {
         const error = new ToolKitError('node test failed to run')
         error.details = err.message
+        error.exitCode = 1
```
ivomurrell, line 133: **issue**: i don't think we want to halt all other tasks if the `NodeTest` task fails – other tasks could complete fine and make progress separately.

ivomurrell (changes requested):
**issue**: it would be better to reexport the `exitCode` [set](https://github.com/Financial-Times/dotcom-tool-kit/blob/c68b55432b3d635203882ec9ef088c5ace6db1ad/lib/logger/src/helpers.ts#L185) by `waitOnExit` rather than overriding it with `1` as often the exit code returned by a failed process is useful for debugging.

**thought**: the pattern of catching errors and re-throwing them with a short message summarising what the task was doing feels a bit redundant. the `runTasks` error handler already [includes](https://github.com/Financial-Times/dotcom-tool-kit/blob/c68b55432b3d635203882ec9ef088c5ace6db1ad/core/cli/src/tasks.ts#L121) the name of the task when aggregating errors (though we should expand that to also include the task name somewhere when [re-throwing](https://github.com/Financial-Times/dotcom-tool-kit/blob/c68b55432b3d635203882ec9ef088c5ace6db1ad/core/cli/src/tasks.ts#L106) errors with an `exitCode` property,) and i don't think custom messages added in the `catch` statements here add much additional context. perhaps it would be okay to not catch the errors in these tasks and let the tool kit core handle them instead, meaning we don't have to worry about losing additional context when creating `ToolKitError`'s here?

plugins/node-test/src/tasks/node-test.ts
rowanmanning: Ah OK, yep that's true 👍 wasn't thinking and just blanket applied it. Will remove

rowanmanning:
> **issue**: it would be better to reexport the `exitCode`...

Good shout, I'll do this.

> **thought**: the pattern of catching errors and re-throwing them with a short message summarising what the task was doing feels a bit redundant...

Maybe I need to try this again but we weren't getting useful error details from thrown `Error` instances (as in not `ToolKitError` ones), which is why we ended up wrapping - there are some things that throw outside of our control, e.g. loading the Hako config file. We _could_ try/catch individual things like this and throw `ToolKitError` explicitly? Or we could adjust the wrapping try/catch to only rethrow the error if it's _not_ an instance of `ToolKitError`:

```js
try {
    // The stuff
} catch (error) {
    if (!(error instanceOf ToolKitError)) {
        // wrap in ToolKitError and throw
    }
    throw error;
}
```

ivomurrell:
> Or we could adjust the wrapping try/catch to only wrap the error if it's not an instance of ToolKitError and throw otherwise

I considered this but tbh I can't see anything in our code that would suggest converting an `Error` to a `ToolKitError` provides additional value (unless you're adding specific contextually-useful messages or `details` of course.) A quick test seems to confirm this – I removed the try/catch block from `node-test` and the only difference when an `Error` (which was not `instanceof ToolKitError`) was thrown was that the stack trace was included in the log (which is a separate issue I should fix with [this logic](https://github.com/Financial-Times/dotcom-tool-kit/blob/c68b55432b3d635203882ec9ef088c5ace6db1ad/core/cli/src/messages.ts#L249) i think) so in reality it gave _more_ context not less 😄

plugins/aws/src/tasks/assume-role.ts
```diff
@@ -43,13 +43,13 @@ export default class AwsAssumeRole extends Task<{ task: typeof AwsAssumeRoleSche
...
       this.logger.info(`Saved AWS credentials to "ci" state with session name "${RoleSessionName}"`)
     } catch (err) {
-      if (err instanceof Error) {
-        const error = new ToolKitError('failed to assume AWS role')
-        error.details = err.message
+      // We wrap non-ToolKitError errors to ensure that they exit the process
+      if (err instanceof Error && !(err instanceof ToolKitError)) {
+        const error = new ToolKitError(err.message)
+        error.exitCode = 1
         throw error
```
ivomurrell, line 50: **nitpick**: sorry this is petty but this line is redundant because we throw on line 52 regardless 😅

ivomurrell (approved):
this is great thank you!!

plugins/aws/src/tasks/assume-role.ts
ivomurrell: ugh yeah sorry i'm running on low sleep today!!

PR #1010, opened 2025-07-07
feat: update zod to 3.25 and explicitly import zod/v3

apaleslimghost:
in #998 i'm going need to update to Zod v4, because how Zod v3 handles defaults doesn't let me detect if the default is being used so i can print a deprecation warning.

starting with Zod v3.25, both the Zod 3.x and 4.x packages export both versions as subpaths. this means we can update to Zod v4 without a breaking change by doing the following:

1. update to Zod v3.25 as a `peerDependency` and import `zod/v3` 👈🏻 you are here
2. update the `peerDependency` range to `^3.25 || 4.x`, switch to importing `zod/v4`, and make any changes to the schemas necessary with `v4`

`core/cli` doesn't rely on plugins using any particular version of Zod; it's just doing `safeParse`, which is compatible between v3 and v4. this PR also updates to `zod-validation-error` v3.5 which is also compatible with both versions of Zod.


apaleslimghost:
discussed with @ivomurrell offline:

- i'm going to switch back to a non-peer dependency on Zod. updating to `4.x` in a plugin (later) won't be a breaking change for that plugin; it will need to update its peer dependency on `dotcom-tool-kit` to require a minimum version of one that supports plugins with v4 schemas
- we'll need to eventually import both `zod-validation-error/v3` and `zod-validation-error/v4`, detect which version a Zod error is from, and format it using the right version. this is the only hard requirement in `core/cli` on plugins having a schema from a particular version of Zod.
- Zod 3.25 is compatible; this is covered by our schema tests and the fact this PR builds

PR #1080, opened 2025-09-26
feat(docker): add build arg for GIT_COMMIT_HASH

alexmuller:
# Description

This allows people to access the commit hash in their applications if they specify:

```
ARG GIT_COMMIT_HASH="dockerfile_default"
ENV GIT_COMMIT_HASH=$GIT_COMMIT_HASH
```

(or similar default value) in their application's Dockerfile.

It's then available as `process.env.GIT_COMMIT_HASH`. 

# Checklist:


plugins/docker/src/tasks/build.ts
```diff
@@ -36,10 +36,14 @@ export default class DockerBuild extends Task<{
             return ['--label', `${label}=${value}`]
           })
 
+        const commitHash = getCommitHash(readState('ci'))
+
         const childBuild = spawn('docker', [
           'buildx',
           'build',
           '--load', // Without this, the image is not stored and so we can't push it later
+          '--build-arg',
```
olusoladavid, line 45: **thought:** I think it would make sense for us to make this fully extensible and make it a task option that can accept as many `build-arg`s as possible. If this is not urgent for @leannecornish-ft , should we consider this direction?

olusoladavid (approved):
Thanks for this quick fix 🫡

PR #1083, opened 2025-09-29
fix: always output requires into .circleci/config.yml

apaleslimghost:
if `requires` is empty, we end up with a `requires: undefined` in the `expectedConfig`, which gets compared against the not-present `requires` key from the current config and fails to match; however, trying to serialise `requires: undefined` to YAML simply does not output the key, so we end up with a situation where Tool Kit (semi-correctly(?)) tells the user to run `dotcom-tool-kit --install`, but that command is a noop.

default `requires` to an empty array so the config parsing and generation are in a consistent state.

### DIFF
--- a/plugins/circleci/src/circleci-config.ts
+++ b/plugins/circleci/src/circleci-config.ts
@@ -470,7 +470,7 @@ const generateWorkflowJobs = (
             } else {
               return requiredName
             }
-          })
+          }) ?? []
         },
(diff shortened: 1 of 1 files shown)


ivomurrell (approved):
I note that this is a bit of a hack because we don't really need to set `requires` to `[]` and instead just could make sure not to set the field if `job.requires` isn't defined, buuuut this would require some horrible ternary spreading probably so yeah why do it like this. In reality we don't really expect any properly-formed jobs to not have any requirements anyway.

